PR #4, opened 2021-08-23
Removes UsageChart component

brunobasto:
This component is too specific to be on a generic component library. We decided to move this to `vip-ui`


djalmaaraujo (approved):
LGTM.

PR #19, opened 2021-09-22
Renames title attribute to label to allow title attributes into DOM

djalmaaraujo:
Breaking change:

This PR renames the `<OptionRow>` `title` attribute to use `label`. This change will allow when using the component users to pass the `title` attribute as a DOM attribute, providing ways of accessibility.

This PR also changes the `label` prop type to `node`. This change allows the user to pass a react element.


brunobasto (approved):
Good move! Thanks @djalmaaraujo

PR #23, opened 2021-10-04
Implement Tooltip using radix to improve accessibility

djalmaaraujo:
# Description

This PR includes `radix-tooltip` as the base for the tooltip component. It also brings the radix dependencies as `dependencies` and not as `devDependencies`.

Using as `devDependencies` would make it hard to keep consistency to all components. 

https://www.radix-ui.com/docs/primitives/components/tooltip

Screen Shot 2021-10-04 at 16 40 42

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,9 @@
 			"name": "@automattic/vip-design-system",
 			"version": "0.7.1",
 			"dependencies": {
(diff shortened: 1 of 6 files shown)


brunobasto (approved):
Tested and LGTM

We've talked about the odd mouseout behavior on slack. @djalmaaraujo is gonna raise that with Radix.

Thank you @djalmaaraujo!

PR #27, opened 2021-11-02
Textarea: use correct htmlFor attribute for Label

mjangda:
Currently we're using `for` which triggers:

> Warning: Invalid DOM property `for`. Did you mean `htmlFor`?

To reproduce:

```
import { Textarea } from '@automattic/vip-design-system';

<Textarea forLabel="my-input" label="Input" /> 
```


brunobasto (approved):
Thank you!

PR #29, opened 2021-11-22
[PIE-2950] Adds React Testing Library with a basic Test for the Avatar component

djalmaaraujo:
Adds:

- React Testing Library 🥳 
- First tests for Avatar component
- Prettier Configuration


brunobasto (approved):
Thank you!

PR #20, opened 2021-09-23
Add/resource list

SaxonF:
Adds 
- a ResourceList that can be used for Audit, Deploys, Backups and any other time based list.
- a generic Time component that formats text and adds appropriate titles

image
image

PR #37, opened 2021-12-06
Add back the Timeline component

djalmaaraujo:
## Description

Some recent changes #20 removed the `Timeline` component. Although `Time` is made for the new styles, some pages are not fully aligned with the North Star VIP design system.

We are putting back the `Timeline` component to keep consistency between some pages.


brunobasto (approved):
Thank you!

PR #51, opened 2022-03-15
PIE-3053 - Add async select component

pandah3:
## Description

Adding an async version of our `SearchSelect` component to allow for paginated results to load asynchronously both in the dropdown selector list and via search.

This allows you to search for options that aren't loaded fully (i.e.- site 115, which is beyond the default 100 queries). You can also scroll to the bottom of the dropdown list to query the next set of paginated results asynchronously:
https://user-images.githubusercontent.com/29221840/159411417-f9beb5c7-5332-4f10-8fea-d8a10489cc49.mov

## Checklist

## Steps to Test

1. Pull down PR.
1. `npm run storybook`
1. Navigate to the component on Storybook (Storybook > Select> Async): http://localhost:6006/?path=/story/select--async
1. Dropdown selection + search should work*

_*This is a very simple rendition of the component. Ideally, you can link to the Dashboard to see it in action (see video above)._

5. Run `npm run test`
6. All tests should pass*

P.S: There's currently a lot of errors due to a semi-related component. Will open a separate PR to address those errors:

```
Warning: Invalid value for prop `cx` on <svg> tag. Either remove it from the element, or pass a string or number value to keep it in the DOM.
```


pandah3:
So many errors 😭 I think most of them were already there since I see them on master too 😭

src/system/Form/Select.test.js
```diff
@@ -0,0 +1,26 @@
...
+  */
+import { Select } from './Select';
+ 
+const defaultProps = {};
+ 
+describe( '<Select />', () => {
+    it( 'renders the Select component with the default placeholder', async () => {
+        const { container } = render( <Select { ...defaultProps } /> );
+
+        const placeholder = screen.getByText( content, element => {
```
brunobasto, line 17: I think the error is on this line: `content` is not defined. It needs a string in order to search an element by text.

brunobasto (changes requested):
Thank you for adding this @pandah3 🎉 

I've also noticed this PR changes `package-lock.json` from version 2 to version 1. This can happen when using an older version of npm. Did you use the version indicated by the `.nvm` file? `package-lock.json` version 2 is supported by `npm@7` and above if I'm not mistaken.

I added an inline comment on the failing test. The other errors seem to be React warnings, but I don't get those when running the tests locally. Maybe that happened due to a lot of dependencies changing due to the  `package-json` version change.

pandah3:
> I've also noticed this PR changes package-lock.json from version 2 to version 1. This can happen when using an older version of npm. Did you use the version indicated by the .nvm file? package-lock.json version 2 is supported by npm@7 and above if I'm not mistaken.

Oof, thank you for catching! I think I changed the `npm` version trying to resolve some issue I had running `npm link` to link the Design System with the Dashboard, but I'm not sure why I downgraded 🤷‍♀️ 

Should be fixed now:
```
-       "lockfileVersion": 1,
+       "lockfileVersion": 2,
```

src/system/Form/Select.test.js
pandah3: Fixed 👍 I was missing some parentheses, but I ended up scrapping this code and just searching for the text itself.

brunobasto (approved):
Tested and looks good! Thank you @pandah3

brunobasto:
I've added a storybook entry so I could see it in action.

pandah3:
>I've added a storybook entry so I could see it in action.

Hmm I added the storybook entry, but I guess it didn't take?
Maybe the commit didn't get pushed 😕 :
```
export const Async = () => {
	let filteredOptions;
	const [ value, setValue ] = useState();

	// We need to return a formatted object for our async wrapper
	const asyncOptions = search => {
		if ( !! search ) {
			filteredOptions = options.filter( ( { label: optionLabel } ) => {
				const matching = optionLabel.toLowerCase().includes( search );

				return matching;
			} );
		}

		const selections = {
			options: filteredOptions ? filteredOptions : options,
			hasMore: true,
		};

		return selections;
	};

	return (
		<Box sx={ { mr: 2, width: 200 } }>
			<Select
				isAsync
				label="Search..."
				value={ value }
				placeholder="Search..."
				loadOptions={ asyncOptions }
				onChange={ newValue => setValue( newValue ) }
			/>
		</Box>
	);
};
```

Thank you for adding either way! I'll see what happened and either update or just keep your example. Thank you!!

PR #60, opened 2022-05-03
Add experimental data attributes to mapped elements

djalmaaraujo:
## Description

Adding some data attributes to some components, so it's easier for other tools like PENDO to match some elements on the screen.

## Checklist

## Steps to Test

1. Pull down PR.
2. `npm run storybook`.
3. Check that http://localhost:6006/?path=/story/wizard--default for each step you can find a `data-order` attribute and one `data-active`
4. Check that http://localhost:6006/?path=/story/optionrow--default in the second example, you can find a `data-order`
5. Check that http://localhost:6006/?path=/story/tabs--default contains `data-active` for the active tab
6.

### DIFF
--- a/src/system/OptionRow/OptionRow.js
+++ b/src/system/OptionRow/OptionRow.js
@@ -24,6 +24,7 @@ const OptionRow = ( {
 	to,
 	small = false,
 	disabled = false,
+	order = null,
 	...props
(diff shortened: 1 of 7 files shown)


brunobasto (approved):
Makes sense. Thank you

PR #63, opened 2022-06-02
Update `<OptionRow />` disabled styles

t-wright:
## Description

Currently when disabled, the `<OptionRow />` component retains some styling which may make it look like it's still enabled. Namely, the background colour change on hover, and the arrow.  This PR removes these styles if the component is disabled. In 91785be65964540f8b33139e78fbd729d410a0e9 I explored further changes that could potentially be applied in a disabled state, but decided that these were maybe too opinionated. 

#### Before
Screen Capture on 2022-06-02 at 13-09-02

#### After
Screen Capture on 2022-06-02 at 13-09-56

## Checklist

## Steps to Test

1. Pull down PR.
2. `npm run build`
1. Copy the `vip-design-system/build` contents into the `vip-ui/node_modules/@automattic/vip-design-system/build` folder
1. Start the vip-ui `npm run dev`
1. Go to http://localhost:3000/apps/2003/production/data
1. View the disabled `<OptionRow />` at at the top of the page and verify hover style and arrow are not displayed.

### DIFF
--- a/src/system/OptionRow/OptionRow.js
+++ b/src/system/OptionRow/OptionRow.js
@@ -107,7 +107,9 @@ const OptionRow = ( {
 				{subTitle && <Text sx={{ mb: 1, color: 'muted' }}>{subTitle}</Text>}
 				{body && <Text sx={{ mb: 0 }}>{body}</Text>}
 			</Box>
-			<Box>{meta ? meta : ! disabled && <MdArrowForward size={24} />}</Box>
+			{ ( false !== meta && '' !== meta ) && (
+				<Box data-testid="meta">{ meta ? meta : <MdArrowForward size={24} /> }</Box>
+			) }
 		</Grid>
 	);
 };
--- a/src/system/OptionRow/OptionRow.js
+++ b/src/system/OptionRow/OptionRow.js
@@ -27,6 +27,8 @@ const OptionRow = ( {
 	disabled = false,
 	order = null,
 	className = null,
+	as,
+	title,
 	...props
 } ) => {
 	const mergedCard = disabled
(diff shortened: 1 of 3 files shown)


mjangda:
Can you add a before + after screenshot or video?

src/system/OptionRow/OptionRow.js
```diff
@@ -107,7 +107,7 @@ const OptionRow = ( {
 				{subTitle && <Text sx={{ mb: 1, color: 'muted' }}>{subTitle}</Text>}
 				{body && <Text sx={{ mb: 0 }}>{body}</Text>}
 			</Box>
-			<Box>{meta ? meta : <MdArrowForward size={24} />}</Box>
+			<Box>{meta ? meta : ! disabled && <MdArrowForward size={24} />}</Box>
```
brunobasto: I wonder if we should not couple the `disabled` prop to the `meta` prop this way. It may be valuable to display the arrow even when the row is disabled - to indicate that if that option was enabled, it would take you somewhere, for example.

brunobasto (approved):
LGTM. Left a minor comment. Thank you!

src/system/OptionRow/OptionRow.js
t-wright: That's a good point. Perhaps it would be better to just allow removing the arrow separately? E.g. by explicitly meta to `false` – `<OptionRow meta={false}  />`? What do you think @brunobasto ?

```suggestion
			{false !== meta && (
				<Box>{meta ? meta : <MdArrowForward size={24} />}</Box>
			)}
```

t-wright: Done in [0b9ae26](https://github.com/Automattic/vip-design-system/pull/63/commits/0b9ae2644b0110f5827261d9aa00a89f3e51698f)

brunobasto (approved):

PR #64, opened 2022-06-08
Upgrade Storybook

chriszarate:
## Description

Upgrade Storybook and addons to 6.x. cc: @maxschmeling since this may have been in your queue.

Notes:

- This moves `preview.js` to a decorator pattern, which is a bit more modular.
- Instead of rendering a toggle button for dark mode, this uses the built in "backgrounds" addon, so toggling dark mode is part of the toolbar.
- I had to add the suggested peer dependencies as devDependencies, but I think this is the right approach? Not sure.
- I also had to rename all stories to `.jsx`, I think Storybook requires this now.

## Steps to Test

1. Pull down PR.
1. `npm i`.
1. `npm run storybook`


djalmaaraujo:
@chriszarate I was able to make it work. `jsdom` is no longer included in JEST 28+.

djalmaaraujo (approved):

PR #70, opened 2022-06-29
Add/alert role to notifications and ScreenReaderText

djalmaaraujo:
## Description

- New proposal for `Notification`.

We are adding the `role="alert"` to the Notification component when it appears. It will not focus on the element.

image

image

- Adding `ScreenReaderText` component for easy screen-reader-text class access

## Checklist

## Steps to Test

1. Pull down PR.
2. `npm run dev`.
3. Activate voice over (cmd+fn+f5 on macOS)
4. Open http://localhost:6006/?path=/story/notification--default
5. You should listen: `Alert, Awesome! Your message has been sent successfully.`

### DIFF
--- a/src/system/Notification/Notification.js
+++ b/src/system/Notification/Notification.js
@@ -11,55 +11,45 @@ import PropTypes from 'prop-types';
  */
 import { Box, Card, Flex, Text, Button } from '../';
 import ScreenReaderText from '../ScreenReaderText/ScreenReaderText';
-import { useLayoutEffect, useRef } from 'react';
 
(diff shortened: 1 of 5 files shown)


brunobasto (approved):
Tested and sounds good to me. Thank you for improving this!

Only minor comment would be about the translation of the word `Alert`, but we can think about that in the future.

🚢

djalmaaraujo:
> Tested and sounds good to me. Thank you for improving this!
> 
> Only minor comment would be about the translation of the word `Alert`, but we can think about that in the future.
> 
> 🚢

Good point. We are not using any internalization here.

PR #74, opened 2022-07-05
Exposes <TableCell> so users have more control over table markups

brunobasto:
## Description

Right now users don't have much control over the markup of the `<td>` on a table. This PR allows them to build tables with more granular control. Like this:

```jsx
<TableRow>
    <TableCell sx={ { backgroundColor: 'gray' } }>
        simon
    </TableCell>
</TableRow>
```

Note that we can now specify styles for the `<td>` or `<th>` which we couldn't do before. One use case is on the Dashboard: I want the first column to have a specific width since it only contains a thumbnail.

***Before***

before_table

***After***

after_table

## Checklist

## Steps to Test

1. Pull down PR.
1. `npm run dev`.
1. Open Storybook.
1. Go to the Table component
1. Verify there's a custom `<td>` with its own background

### DIFF
--- a/src/system/Table/Table.stories.jsx
+++ b/src/system/Table/Table.stories.jsx
@@ -1,7 +1,7 @@
 /**
  * Internal dependencies
  */
-import { Table, TableRow, Flex, Heading, Text } from '..';
+import { Table, TableRow, Flex, Heading, Text, TableCell } from '..';
 
 export default {
 	title: 'Table',
@@ -36,6 +36,34 @@ export const Default = () => (
 					</Text>,
 				] }
 			/>
+			<TableRow>
+				<TableCell sx={ { backgroundColor: 'gray' } }>
+					<Flex sx={ { alignItems: 'center' } } key="user">
+						<Heading variant="h4" sx={ { mb: 0 } }>
+							simon
+						</Heading>
+					</Flex>
+				</TableCell>
+				<TableCell>
+					<Heading
+						variant="h4"
+						key="command"
+						sx={ { mb: 0, display: 'flex', alignItems: 'center' } }
+					>
+						wp posts list
+					</Heading>
+				</TableCell>
+				<TableCell>
+					<Text sx={ { mb: 0 } } key="duration">
+						3s
+					</Text>
+				</TableCell>
+				<TableCell>
+					<Text sx={ { mb: 0, color: 'muted' } } key="time">
+						3rd May 2021, 13:22:13
+					</Text>
+				</TableCell>
+			</TableRow>
 		</tbody>
 	</Table>
 );
--- a/src/system/Table/Table.stories.jsx
+++ b/src/system/Table/Table.stories.jsx
@@ -37,7 +37,7 @@ export const Default = () => (
 				] }
 			/>
 			<TableRow>
-				<TableCell sx={ { backgroundColor: 'gray' } }>
+				<TableCell sx={ { backgroundColor: 'lightgray' } }>
 					<Flex sx={ { alignItems: 'center' } } key="user">
 						<Heading variant="h4" sx={ { mb: 0 } }>
 							simon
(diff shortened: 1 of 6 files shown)


src/system/Table/TableCell.js
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Internal dependencies
+ */
+import { Heading, Box } from '../';
+
+const TableCell = ( { head, isRowHead, children, ...rest } ) => {
+	let scope = null;
+
+	if ( head ) {
```
andrea-sdl: @brunobasto unrelated to this PR but I find the "head/isRowHead" a little bit confusing. Even by looking at the code I always wonder what they are used for.

I'm sure renaming them might not be the case now but I was wondering if we could add some extra example in the stories to clarify the use of the parameters

src/system/Time/index.js
```diff
@@ -56,7 +56,7 @@ const Time = ( { time, relativeTime = false, timeOnly = false, className = null,
 };
 
 Time.propTypes = {
-	time: PropTypes.oneOfType( [ PropTypes.string, PropTypes.date ] ),
+	time: PropTypes.oneOfType( [ PropTypes.string, PropTypes.instanceOf( Date ) ] ),
```
andrea-sdl, line 59: @brunobasto didn't see it mentioned in the PR, is it a leftover or not? (I believe the new code is correct so I don't consider this an issue, just wondering)

brunobasto: Oh, forgot to mention on the PR description, but mentioned on the commit message:

(https://github.com/Automattic/vip-design-system/pull/74/commits/5603609fc5280f83cdda433183171daf6fae5689) 
```
Fix invalid PropType error:

`Warning: Invalid argument supplied to oneOfType. Expected an array of check functions, but received undefined at index 1.`
```

Noticed this error when running tests.

src/system/Table/TableCell.js
brunobasto: I agree it's confusing. I'll play with the storybook and add some examples.

src/system/Table/Table.stories.jsx
```diff
@@ -36,6 +36,34 @@ export const Default = () => (
 					</Text>,
 				] }
 			/>
+			<TableRow>
+				<TableCell sx={ { backgroundColor: 'gray' } }>
```
andrea-sdl:
```suggestion
				<TableCell sx={ { backgroundColor: 'lightgray' } }>
```
Moving it to `lightgray` still allows this to be visible and doesn't raise an A11Y violation in the storybook

andrea-sdl (approved):
LGTM, tested and it works 🚢 

Only a couple of improvements which are optional left as comments

src/system/Table/Table.stories.jsx
brunobasto: Awesome! I didn't notice the violation. Thank you

src/system/Table/TableCell.js
brunobasto: So, essentially, `head` and `isRowHead` control two things:

1. Whether to render the columns as `<td>` or `<th>`. If either is true, the column will render as `<th>`
2. The scope of the column header:

> The scope attribute specifies whether a header cell is a header for a column, row, or group of columns or rows.
> 
> **Note**: The scope attribute has no visual effect in ordinary web browsers, but can be used by screen readers. 
> 
> https://www.w3schools.com/tags/att_th_scope.asp

I searched in `vip-ui` and couldn't find any usages of `rowHead` or `isRowHead`, so, for the sake of simplicity in the API, I [removed](https://github.com/Automattic/vip-design-system/pull/74/commits/e5e87840180bc9aa4f23e65609dd963d0a117fe3) `isRowHead` and, since we pass forward the props, if someone ever needs to control the scope for accessibility purposes, they can do:

```jsx
<TableCell head scope="row">...
```

brunobasto: Since the prop I removed was unused (so it shouldn't impact the Dashboard) and the PR is already approved, I'll be bold and move forward with merging this, but we can add it back quickly if we need to.

PR #77, opened 2022-07-11
[PIE-3193] New Dialog using Radix component

djalmaaraujo:
## Description

This PR introduces the new `Dialog` component. It's rebuilt from zero using the Radix Dialog component. The main goal is to keep consistency with the existing features and props while improving the accessibility of the whole component.

Our old component did not have the proper a11y capabilities, such as:

- Focus management (when open / close)
- Expandable aria attributes
- A "Close button" with proper labels
- Accessible title and descriptions for screenreaders

This new implementation includes all a11y benefits of the Radix Dialog, keeps the API consistent, and adds the VIP theme at the top of it.

### Reference
- [Radix Dialog Documentation](https://www.radix-ui.com/docs/primitives/components/dialog)

### Research
Before choosing Radix, I also researched the best way to implement dialogs. Here we have some links if other wants to read more about it:

- [Dialog (Simple Dialog)](https://dequeuniversity.com/library/aria/simple-dialog)
- [✎ Technique: Accessible modal dialogs | Digital Accessibility​](https://accessibility.huit.harvard.edu/technique-accessible-modal-dialogs)
- [Modal Dialog Example | WAI-ARIA Authoring Practices 1.1](https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html)
- [The Dialog element - HTML: HyperText Markup Language | MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog)
- https://a11y-dialog.netlify.app/

## Checklist

## Screenshots

image

image

image

**Ps**: The dialogs are not aligned to the bottom, I cropped the screenshots because my monitor is too big 😆 .

## Steps to Test

1. https://deploy-preview-77--vip-design-system-components.netlify.app/?path=/story/newdialog--default
2. https://deploy-preview-77--vip-design-system-components.netlify.app/?path=/story/newdialog--auto-open
3. https://deploy-preview-77--vip-design-system-components.netlify.app/?path=/story/newdialog--hidden-headings
4. https://github.com/Automattic/vip-design-system/pull/77


rianrietveld:
@djalmaaraujo another issues to take into account:
 If a modal contains more than just a form, it needs a role="document" to announce the extra text.
 
A screen reader jumps into "form mode" entering a dialog, only targeting focusable elements.
But if there is also additional info, outside a form, a screenreader should handle that in document mode.
It's all nicely written down in:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/document_role

But, if you link the text with aria-labelledby and aria-described by to the dialog element, this is not needed.
It depends on the complexity of the dialog.

brunobasto (approved):
🚢 🚢 🚢

djalmaaraujo:
@rianrietveld Just letting you know. I am merging this first version to start using without the role="document" in mind, and we will see in the subsequent iterations if we need it.

PR #78, opened 2022-07-14
Add another visual indicator for the Toggle component state

djalmaaraujo:
## Description

According to the Guidelines ([this one as a reference](https://developer.wordpress.org/block-editor/reference-guides/components/form-toggle/)) and our expert consultant @rianrietveld, it's a WCAG requirement to have an indicator of the toggle state that is not only a color indicator. For folks with color blindness, for example, the color would not matter.

We are adding a visual indicator similar to the WordPress [FormToggle component](https://developer.wordpress.org/block-editor/reference-guides/components/form-toggle/) indicating when the Toggle is On or Off.

image

Screen Shot 2022-07-15 at 14 26 14

Ps: Bonus: You can pass a `variant` to determine which theme color you want. Any color available on the theme is possible.

## Checklist

## Steps to Test

1. Pull down PR.
2. `npm run dev`.
3. Open https://deploy-preview-78--vip-design-system-components.netlify.app/?path=/story/toggle--primary
4. You should see the symbol "✓" as an indicator of ON
5. You should see the symbol "✕" as an indicator of OFF

### DIFF
--- a/src/system/Form/Toggle.js
+++ b/src/system/Form/Toggle.js
@@ -23,20 +23,21 @@ export const Toggle = ( { name = 'toggle', onChange, className = null, ...rest }
 			position: 'relative',
 			WebkitTapHighlightColor: 'rgba(0, 0, 0, 0)',
 			'::after': {
-				content: '"o"',
+				fontFamily: 'system-ui, Helvetica, sans-serif',
+				content: '"\\2715"',
 				position: 'absolute',
 				color: 'white',
 				fontSize: '10px',
 				fontWeight: 'bold',
 				top: 1,
-				right: 10,
+				right: '9px',
 			},
 			'&[data-state="checked"]': {
 				backgroundColor: 'success',
 				'::after': {
-					content: '"I"',
+					content: '"\\2713"',
 					top: 1,
-					left: 10,
+					left: '8px',
 				},
 			},
 		} }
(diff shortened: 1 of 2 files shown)


GaryJones:
Semantically, capital I and lowercase o (Not enter uppercase O) don't mean much.

It's also relying on a particular font to make the I glyph look like a | bar, and users may not have access to that font, or they may be overriding it for their own accessibility benefits.

djalmaaraujo:
> Semantically, capital I and lowercase o (Not enter uppercase O) don't mean much.
> 
> It's also relying on a particular font to make the I glyph look like a | bar, and users may not have access to that font, or they may be overriding it for their own accessibility benefits.

Do you have any suggestions on what font should be used? Do you suggest using a Pipe (|) and an O? The idea is to give a visual difference, I don't see how we could show any other information to the user. We are not adding any readable text since the toggle itself should be clear. On/Off could be an option but also lacks context depending on what you are putting on/off or activating/deactivating.

Do you have any suggestions?

WPprodigy:
Checkmark (BiCheck icon maybe ) and an `X` symbol are the most "clear" and flexible for any situation I think.

download

djalmaaraujo:
@GaryJones @WPprodigy Updated with a standard check/x CSS chars. Also, I applied the font to be system-ui, so it would get clear for all systems. Let me know what you think now.

Looks way better from my perspective.

GaryJones (approved):

brunobasto (approved):
Looks great! Thank you all

rianrietveld:
nice!

PR #84, opened 2022-07-21
Add/toggle hover dialog document

djalmaaraujo:
## Description

- Adds a cursor pointer to the Toggle component (Required to have a hover state);
- Adds a `role="document"` attribute to the Dialog content. This is to avoid having developers to remember to add this `role` for Dialogs with more than just a form.

## Checklist

## Steps to Test — Toggle

1. Pull down PR.
2. `npm run dev`.
3. Open the [Toggle Example](http://localhost:6006/?path=/story/toggle--primary)
4. When HOVERING the Toggle you should see a cursor pointer (hand)

## Steps to Test — Dialog

1. Pull down PR.
2. `npm run dev`.
3. Open the [NewDialog Example](http://localhost:6006/?path=/story/newdialog--default)
4. Inspect the Dialog. The content wrapper div should have a `role="document"` for all dialogs.

### DIFF
--- a/src/system/NewDialog/NewDialog.js
+++ b/src/system/NewDialog/NewDialog.js
@@ -56,7 +56,7 @@ export const NewDialog = ( {
 					<DialogTitle title={ title } hidden={ ! showHeading } />
 					<DialogDescription description={ description } hidden={ ! showHeading } />
 
-					{ content }
+					<div role="document">{ content }</div>
 				</DialogPrimitive.Content>
 			</DialogPrimitive.Portal>
 		</DialogPrimitive.Root>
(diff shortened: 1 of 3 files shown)


src/system/NewDialog/NewDialog.js
```diff
@@ -56,7 +56,7 @@ export const NewDialog = ( {
 					<DialogTitle title={ title } hidden={ ! showHeading } />
 					<DialogDescription description={ description } hidden={ ! showHeading } />
 
-					{ content }
+					<div role="document">{ content }</div>
```
brunobasto, line 59: Out of curiosity, why don't we use role=dialog? 

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role

djalmaaraujo: We already have that in the dialog. This is just for the content.

brunobasto: oh, I see. Thanks for clarifying 🙇

brunobasto (approved):

PR #88, opened 2022-07-25
Fix/new dialog references

### LINKED ISSUE #86: NewDialog doesn't return focus correctly on close
As per https://www.w3.org/WAI/ARIA/apg/example-index/dialog-modal/dialog when a dialog is closed, such as via pressing Escape or clicking on an X, the focus should go back to the button or other element that triggered it. This is not currently the case on https://vip-design-system-components.netlify.app/?path=/story/newdialog--default (or other variants) which seems to put the focus back on the whole Storybook window.

Steps to demonstrate:

1. Go to https://www.w3.org/WAI/ARIA/apg/example-index/dialog-modal/dialog
2. Click the Add Delivery Address button. A dialog will appear.
3. Press Escape, and the dialog will close, and the focus will be back on the button.
4. Hitting Enter will re-open the dialog again immediately.
5.  Now visit https://vip-design-system-components.netlify.app/?path=/story/newdialog--default
6. Click the Trigger Dialog button, and the dialog will open.
7. Hit Escape (or Enter, since the focus is already on the X), and the dialog will close, but focus is not back on the button (no blue border).
8. Hit Enter again, but nothing will happen. Hit `s` (as. Storybook shortcut) will toggle the Storybook sidebar, showing focus is on there. Hitting Shift-Tab is needed to get focus back on to the Trigger Dialog button.

djalmaaraujo:
## Description

- Fixes #86 

- This PR fixes the issue when the Dialog closes the focus doesn't get back to the previous trigger. The problem is that some of our components don't support passing a `ref` attribute. This PR fixes the Button reference and adds support to export the `Close` primitive container so that the VIP Dashboard can implement custom Close buttons inside of the Dialog.

- Radix Dialog upgraded to 1.0.0

## Checklist

## Steps to Test — Focus fix

- Follow the steps on issue #86 

## Steps to Test - Custom Close

- Go to http://localhost:6006/?path=/story/newdialog--custom-close and use the internal close button. The focus should stay on the original trigger.


luiztiago (approved):
LGTM :shipit:

GaryJones (approved):
I can't attest to the quality of the code changes, but all variants of [NewDialog](https://deploy-preview-88--vip-design-system-components.netlify.app/?path=/story/newdialog--default) now correctly give focus back to the trigger button if it exists.✅

PR #94, opened 2022-08-10
Adds new NewConfirmationDialog component 🥳 

djalmaaraujo:
## Description

aEMfVx36HP

Adding a new component: src/system/NewConfirmationDialog/NewConfirmationDialog.js

This is a new version of the `ConfirmationDialog`. Following the same idea as the `NewDialog`, we are creating a new component to avoid breaking the clients of the vip-design-system.

## Checklist

## Steps to Test

1. Pull down PR.
2. `npm run dev`.
3. Go to http://localhost:6006/?path=/story/newconfirmationdialog--default
4. Verify the Dialog works and it has the same accessibility as the NewDialog component. Focus is on the Close button, user can navigate using the keyboard, can close using ESC key, and focus is back to the trigger.

### DIFF
--- a/src/system/NewConfirmationDialog/NewConfirmationDialog.js
+++ b/src/system/NewConfirmationDialog/NewConfirmationDialog.js
@@ -1,0 +1,93 @@
+/** @jsxImportSource theme-ui */
+
+/**
+ * External dependencies
+ */
+import classNames from 'classnames';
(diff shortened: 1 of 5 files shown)


brunobasto (approved):
Tested and works as expected! Thank you @djalmaaraujo

PR #96, opened 2022-08-12
Add onClose exported to NewDialog Content

djalmaaraujo:
## Description

The current implementation of our Dialog has this feature where you can close the dialog using a function on the Dialog content property.

You can check the StoryBook example in the changed files to understand better.

## Checklist

## Steps to Test

1. Pull down PR.
2. `npm run dev`.
3. Open https://deploy-preview-96--vip-design-system-components.netlify.app/?path=/story/newdialog--custom-on-close
4. Check the example works as expected, controlling the dialog using the exported `onClose`.

### DIFF
--- a/src/system/NewDialog/NewDialog.js
+++ b/src/system/NewDialog/NewDialog.js
@@ -31,7 +31,7 @@ export const NewDialog = ( {
 	open = undefined,
 	id = undefined,
 } ) => {
-	const [ isOpen, setIsOpen ] = React.useState( defaultOpen );
+	const [ isOpen, setIsOpen ] = React.useState( open || defaultOpen );
 
 	if ( disabled ) {
 		return;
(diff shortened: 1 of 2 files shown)


src/system/NewDialog/NewDialog.js
```diff
@@ -30,15 +31,20 @@ export const NewDialog = ( {
...
 
+	// if content is a function, pass in onClose
+	const isContentFunction = typeof content === 'function';
+
 	return (
 		<DialogPrimitive.Root
 			id={ id }
-			open={ open }
-			onOpenChange={ onOpenChange }
+			open={ open || isOpen }
```
brunobasto: Can this be problematic? If the initial prop value for `open` is `true` and doesn't change, than the dialog will always stay open?

djalmaaraujo: @brunobasto You would need to customize the `onOpenChange`, yes. Do you have a suggestion for this case?

brunobasto (approved):
Looks great! Thank you

PR #100, opened 2022-08-24
Adds new component: Dropdown

djalmaaraujo:
## Description

Adding the `Dropdown` new component 🥳 . This component should replace all `DialogMenu` with all his children: `DialogMenuItem`, etc.

CvH80UXKrf

The new component provides:

- [Full documented API from Radix](https://www.radix-ui.com/docs/primitives/components/dropdown-menu)
- A11Y capabilities
- Dark mode support
- Support to Sub menus
- Support to Checkbox items
- Support to open a Modal and keep focus after the Dialog is closed (pending Story)

## Checklist

## Steps to Test

1. Pull down PR.
2. `npm run dev`.
3. Open Storybook.
4. Check how the Dropdown behaves similarly to the <Dialog> component, but with accessibility. You can use a keyboard, and navigate using sub-menus, etc.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -40,6 +40,7 @@
 				"@babel/preset-react": "^7.12.13",
 				"@babel/runtime": "7.14.0",
 				"@mdx-js/react": "^2.1.1",
+				"@radix-ui/react-icons": "^1.1.1",
 				"@storybook/addon-a11y": "^6.5.9",
 				"@storybook/addon-actions": "^6.5.9",
 				"@storybook/addon-controls": "^6.5.9",
@@ -5200,6 +5201,15 @@
 			},
 			"peerDependencies": {
 				"react": "^16.8 || ^17.0 || ^18.0"
+			}
+		},
+		"node_modules/@radix-ui/react-icons": {
+			"version": "1.1.1",
+			"resolved": "https://registry.npmjs.org/@radix-ui/react-icons/-/react-icons-1.1.1.tgz",
+			"integrity": "sha512-xc3wQC59rsFylVbSusQCrrM+6695ppF730Q6yqzhRdqDcRNWIm2R6ngpzBoSOQMcwnq4p805F+Gr7xo4fmtN1A==",
+			"dev": true,
+			"peerDependencies": {
+				"react": "^16.x || ^17.x || ^18.x"
 			}
 		},
 		"node_modules/@radix-ui/react-radio-group": {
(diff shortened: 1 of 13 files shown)


brunobasto (approved):

rianrietveld:
@djalmaaraujo 
Thanks! This works much better.
I have a few remarks.

Add an aria-expanded="false" toggle on the button. Not it is only added when the aria-expanded="true".
`<button class="vip-button-component css-1qlg2ts" type="button" id="radix-0" aria-haspopup="menu" data-state="closed" aria-expanded="false">Open</button>`

Indicate the active item not by a slight changes of color alone, the color changes from black to brown, add an other way to show the active item for color blind users. A common way is to invert the text and background colors, or an underlining or bold text, (I am not a designer).

A usability issue, not a blocker: You can not return to then button with keyboard arrowing up, only with Esc. It's a mix between a modal and a dropdown actually.

rianrietveld:
@djalmaaraujo
About the Complex options
The "More tools" options isn't really understandable for screen reader users, if you don't see what is happening.
There is no way to return in an initiative way back to the More tools option if you entered the new opened options.
The ticked Show Bookmarks, what does that mean? It announced as "Ticked," but is it selected? The current option?

The word "People" is not announced by a screenreader. Only the name (and in this case) that Pedro Duarte is ticked.

I highly recommend not to use this Complex Options version, because it will be very complex to make this accessible and announce well.

An alternative could be a normal menu-dropdown construction when using an option like "More Tools"
And a normal HTML5 select when you want to group options like with People here.
See [<optgroup>: The Option Group element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup) on MDN.

djalmaaraujo:
> @djalmaaraujo Thanks! This works much better. I have a few remarks.
> 
> Add an aria-expanded="false" toggle on the button. Not it is only added when the aria-expanded="true". `<button class="vip-button-component css-1qlg2ts" type="button" id="radix-0" aria-haspopup="menu" data-state="closed" aria-expanded="false">Open</button>`
> 
> Indicate the active item not by a slight changes of color alone, the color changes from black to brown, add an other way to show the active item for color blind users. A common way is to invert the text and background colors, or an underlining or bold text, (I am not a designer).
> 
> A usability issue, not a blocker: You can not return to then button with keyboard arrowing up, only with Esc. It's a mix between a modal and a dropdown actually.

It seems to be a Radix issue, I opened an issue: https://github.com/radix-ui/primitives/issues/1633.

Thanks for the report @rianrietveld

PR #103, opened 2022-09-06
Adds new component: NewForm/Select

luiztiago:
## Description

Adding the `NewForm/Select` new component 🥳 . 

Select Default

Select with Focus

Select Default with options visible

Select Default with Grouped options

Screen Shot 2022-09-08 at 17 57 57

Screen Shot 2022-09-08 at 17 58 03

## Checklist

## Steps to Test

1. Pull down PR.
1. `npm run dev`.
1. Open Storybook - https://deploy-preview-103--vip-design-system-components.netlify.app/?path=/story/newform-select--primary
1. Check how the Select behaves similarly to the component, but with accessibility. You can use a keyboard, and navigate between options, etc.


rianrietveld:
@djalmaaraujo 
The color contrast of the select control borders and icons must be 3:1 or higher.

Both form controls work nicely with keyboard and VoiceOver in Safari.
The group is announced as dimmed, this is the same as for the native HTML select behaviour.

About your question: the new Select component does not open with the "RETURN" key pressed, the same as the default browser solution. Radix` <Select>` opens with RETURN, what should we do?  

The radix select is in fact a button that opens a sort of dialog. It's announced as a button, so should open with return (and spacebar). This is expected behaviour.

djalmaaraujo:
> @djalmaaraujo The color contrast of the select control borders and icons must be 3:1 or higher.
> 
> Both form controls work nicely with keyboard and VoiceOver in Safari. The group is announced as dimmed, this is the same as for the native HTML select behaviour.
> 
> About your question: the new Select component does not open with the "RETURN" key pressed, the same as the default browser solution. Radix` <Select>` opens with RETURN, what should we do?
> 
> The radix select is in fact a button that opens a sort of dialog. It's announced as a button, so should open with return (and spacebar). This is expected behaviour.

Thanks for the comments. Design team is working on the new colors, so it should be fixed as soon as I get a new update from them.

brunobasto (approved):

PR #119, opened 2022-09-23
Fixed Form.Select options using getAllOptions

luiztiago:
## Description

Fixed onChange returning the expected option when using grouped options and when the current value is not a string on new Select component 🥳 . 

image

## Checklist

## Steps to Test

1. Pull down PR.
1. `npm run dev`.
1. Open Storybook - https://deploy-preview-119--vip-design-system-components.netlify.app/?path=/story/form-select--with-on-change
1. Check the entire object returned on changing a different option in select.

### DIFF
--- a/src/system/NewForm/FormSelect.js
+++ b/src/system/NewForm/FormSelect.js
@@ -3,7 +3,7 @@
 /**
  * External dependencies
  */
-import React, { useCallback } from 'react';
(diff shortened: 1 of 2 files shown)


djalmaaraujo (approved):
Works as expected. Good catch on the === comparison.

PR #126, opened 2022-10-26
PIE-3411 Adds Accordion component

brunobasto:
## Description

Adds the accordion component to be used on the Prep Panel side bar.

4IeiWfhVi3

## Checklist

## Steps to Test

1. Pull down PR.
1. `npm run dev`.
1. Open Storybook.
1. Go to the Accordion component

### DIFF
--- a/src/system/Accordion/Accordion.js
+++ b/src/system/Accordion/Accordion.js
@@ -72,7 +72,7 @@ export const Trigger = React.forwardRef(
 					cursor: 'pointer',
 					all: 'unset',
 					fontFamily: 'inherit',
-					padding: '0 20px',
+					px: 3,
 					height: 45,
 					flex: 1,
 					display: 'flex',
@@ -117,6 +117,20 @@ Trigger.propTypes = {
 	headingVariant: PropTypes.string,
 };
 
+export const TriggerWithIcon = React.forwardRef( ( { children, icon, ...props }, forwardedRef ) => (
+	<Trigger { ...props } ref={ forwardedRef }>
+		<span sx={ { color: theme => theme.colors.grey[ '70' ], fontSize: 20 } }>{ icon }</span>
+		<div sx={ { flexGrow: 1, textAlign: 'left', ml: 3 } }>{ children }</div>
+	</Trigger>
+) );
+
+TriggerWithIcon.displayName = 'Accordion.TriggerWithIcon';
+
+TriggerWithIcon.propTypes = {
+	children: PropTypes.node.isRequired,
+	icon: PropTypes.node.isRequired,
+};
+
 export const Content = React.forwardRef( ( { children, ...props }, forwardedRef ) => {
 	return (
 		<AccordionPrimitive.Content
@@ -138,7 +152,8 @@ export const Content = React.forwardRef( ( { children, ...props }, forwardedRef 
 		>
 			<div
 				sx={ {
-					padding: '15px 20px',
+					px: 3,
+					py: 2,
 				} }
 			>
 				{ children }
--- a/src/system/Accordion/Accordion.js
+++ b/src/system/Accordion/Accordion.js
@@ -1,0 +1,184 @@
+/** @jsxImportSource theme-ui */
+
+/**
+ * External dependencies
+ */
+import React from 'react';
+import PropTypes from 'prop-types';
+import { MdChevronRight } from 'react-icons/md';
+import * as AccordionPrimitive from '@radix-ui/react-accordion';
+import { keyframes } from '@emotion/react';
+
+/**
+ * Internal dependencies
+ */
+import { Heading } from '../Heading';
+
+const slideDown = keyframes( {
+	from: { height: 0 },
+	to: { height: 'var(--radix-accordion-content-height)' },
+} );
+
+const slideUp = keyframes( {
+	from: { height: 'var(--radix-accordion-content-height)' },
+	to: { height: 0 },
+} );
+
+export const Item = ( { children, ...props } ) => (
+	<AccordionPrimitive.Item
+		{ ...props }
+		sx={ {
+			overflow: 'hidden',
+			borderWidth: '0 1px 1px 1px',
+			borderStyle: 'solid',
+			borderColor: theme => theme.colors.gray[ '7' ],
+
+			'&:first-of-type': {
+				borderTopWidth: '1px',
+				borderTopLeftRadius: 4,
+				borderTopRightRadius: 4,
+			},
+			'&:last-child': {
+				borderBottomLeftRadius: 4,
+				borderBottomRightRadius: 4,
+			},
+			'&:focus-within': {
+				zIndex: 1,
+				outline: 'auto',
+			},
+		} }
+	>
+		{ children }
+	</AccordionPrimitive.Item>
+);
+
+Item.displayName = 'Accordion.Item';
+
+Item.propTypes = {
+	children: PropTypes.node.isRequired,
+};
+
+export const Trigger = React.forwardRef(
+	( { children, headingVariant = 'h3', ...props }, forwardedRef ) => (
+		<Heading
+			sx={ {
+				all: 'unset',
+				display: 'flex',
+			} }
+			variant={ headingVariant }
+		>
+			<AccordionPrimitive.Trigger
+				sx={ {
+					cursor: 'pointer',
+					all: 'unset',
+					fontFamily: 'inherit',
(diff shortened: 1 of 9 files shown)


src/system/theme/index.js
```diff
@@ -90,6 +90,8 @@ export default {
 		text: getColor( 'text', 'secondary' ),
 		heading: getColor( 'text', 'primary' ),
 		background: getColor( 'background', 'primary' ),
+		gold: getVariants( 'color.gold' ),
+		gray: getVariants( 'color.gray' ),
```
djalmaaraujo, line 94: Isn't gold the same as primary?

djalmaaraujo (approved):
@brunobasto Left one comment regarding colors. Usually, borders are `border` for inputs and borders.2 for fancy borders.

Another thing I think we can do is to have a simpler API for the accordion later? Like:

```
<Accordion content=[{
  id: 'teamPermissions', // or we generate it dynamically
  title: 'Team & Permissions',
  icon: <Icon />,
  content: <div />,
},
{
  id: 'abc', // or we generate it dynamically
  title: 'VYZ',
  icon: <Icon />,
  content: <div />,
}
] />
```

This would be a more straightforward API, but the other one would provide more flexibility.

What do you think?

src/system/theme/index.js
djalmaaraujo: We want to avoid pointing colors from their names, but more like semantics: Primary, Border, Background, etc.

src/system/Accordion/Accordion.js
```diff
@@ -0,0 +1,199 @@
...
+				borderTopLeftRadius: 4,
+				borderTopRightRadius: 4,
+			},
+			'&:last-child': {
+				borderBottomLeftRadius: 4,
+				borderBottomRightRadius: 4,
+			},
+			'&:focus-within': {
+				zIndex: 1,
+				outline: 'auto',
```
djalmaaraujo: I think you can use `outline` from the Theme file. There's a style predefined

src/system/Accordion/Accordion.js
```diff
@@ -0,0 +1,199 @@
...
+					fontFamily: 'inherit',
+					px: 3,
+					height: 45,
+					flex: 1,
+					display: 'flex',
+					alignItems: 'center',
+					justifyContent: 'space-between',
+					fontSize: 12,
+					fontWeight: 600,
+					lineHeight: '14px',
```
djalmaaraujo: Can we use 1 or 2? Looks like theme-ui would fit

src/system/Accordion/Accordion.js
```diff
@@ -0,0 +1,199 @@
...
+					textTransform: 'uppercase',
+					color: theme => theme.colors.black,
```
djalmaaraujo: Instead of black, would it be heading, or text?

src/system/Accordion/Accordion.js
```diff
@@ -0,0 +1,199 @@
...
+
+					'&[data-state="closed"]': { backgroundColor: 'transparent' },
+					'&[data-state="open"]': {
+						backgroundColor: theme => theme.colors.gold[ '7' ],
+						borderBottom: theme => `1px solid ${ theme.colors.gray[ '7' ] }`,
```
djalmaaraujo: I think borders are just borders. That's the default we use. Try `border` or `borders.2`

djalmaaraujo (changes requested):
@brunobasto Left minor comments. AWESOME Work.

Another thing I think we can do is to have a simpler API for the accordion later? Like:

<Accordion content=[{
  id: 'teamPermissions', // or we generate it dynamically
  title: 'Team & Permissions',
  content: <div />,
},
{
  id: 'abc', // or we generate it dynamically
  title: 'VYZ',
  content: <div />,
}
] />
This would be a more straightforward API, but the other one would provide more flexibility.

What do you think?

src/system/Accordion/Accordion.js
```diff
@@ -0,0 +1,199 @@
...
+					cursor: 'pointer',
+					all: 'unset',
+					fontFamily: 'inherit',
+					px: 3,
+					height: 45,
+					flex: 1,
+					display: 'flex',
+					alignItems: 'center',
+					justifyContent: 'space-between',
+					fontSize: 12,
```
djalmaaraujo: I think 1 as a value here would be 12?

src/system/Accordion/Accordion.js
```diff
@@ -0,0 +1,199 @@
...
+	icon: PropTypes.node.isRequired,
+};
+
+export const Content = React.forwardRef( ( { children, ...props }, forwardedRef ) => {
+	return (
+		<AccordionPrimitive.Content
+			sx={ {
+				backgroundColor: 'white',
+				color: theme => theme.colors.text,
+				fontSize: 15,
```
djalmaaraujo: I don't think 15 is a standard size, can you try 1?

src/system/Accordion/Accordion.js
```diff
@@ -0,0 +1,199 @@
...
+	children: PropTypes.node.isRequired,
+	icon: PropTypes.node.isRequired,
+};
+
+export const Content = React.forwardRef( ( { children, ...props }, forwardedRef ) => {
+	return (
+		<AccordionPrimitive.Content
+			sx={ {
+				backgroundColor: 'white',
+				color: theme => theme.colors.text,
```
djalmaaraujo: `color: 'text'` did not work?

src/system/Accordion/Accordion.js
```diff
@@ -0,0 +1,199 @@
...
+Content.displayName = 'Accordion.Content';
+
+Content.propTypes = {
+	children: PropTypes.node.isRequired,
+};
+
+const Root = React.forwardRef(
+	( { sx = {}, defaultValue, type, children, className }, forwardRef ) => (
+		<AccordionPrimitive.Root
+			className={ className }
```
djalmaaraujo: Maybe we should use the default class name for the component?

Also, we can combine classes:

https://github.com/Automattic/vip-design-system/blob/trunk/src/system/Button/Button.js#L33

src/system/Accordion/Accordion.js
```diff
@@ -0,0 +1,199 @@
...
+} );
+
+export const Item = ( { children, ...props } ) => (
+	<AccordionPrimitive.Item
+		{ ...props }
+		sx={ {
+			overflow: 'hidden',
+			borderWidth: '0 1px 1px 1px',
+			borderStyle: 'solid',
+			borderColor: theme => theme.colors.gray[ '7' ],
```
djalmaaraujo: I think this should be or borders or borders.2. I think the first ,since it needs the same contrast as the inputs.

src/system/Accordion/Accordion.js
```diff
@@ -0,0 +1,199 @@
...
+				borderTopWidth: '1px',
+				borderTopLeftRadius: 4,
+				borderTopRightRadius: 4,
+			},
+			'&:last-child': {
+				borderBottomLeftRadius: 4,
+				borderBottomRightRadius: 4,
+			},
+			'&:focus-within': {
+				zIndex: 1,
```
djalmaaraujo: Is this needed?

src/system/theme/index.js
brunobasto: Fixed here: https://github.com/Automattic/vip-design-system/pull/126/commits/6f25718548361d93a84ae08bcafd72a92ae384ff

brunobasto: Gold is not the same as primary. This is the gold referenced in Figma: https://github.com/Automattic/vip-design-system/blob/trunk/tokens/valet-core.json#L782

src/system/Accordion/Accordion.js
brunobasto: Not needed. That was a leftover from the example. Thanks for catching this.

src/system/Accordion/Accordion.js
brunobasto: Good one. Added https://github.com/Automattic/vip-design-system/pull/126/commits/ce4685fb5d8416d1f76102a9f39ba5a35592caa5

brunobasto:
> Another thing I think we can do is to have a simpler API for the accordion later?

100% Agree. Was planning to do this later on once I have a better idea of how things will fit together.

djalmaaraujo (approved):
This looks amazing @brunobasto. Awesome work. Thank you!

PR #130, opened 2022-11-08
Add accessible tabs component

andrea-sdl:
## Description
Screenshot 2022-11-08 at 15 59 57

Introducing a more accessible Tab component by using the Radix one https://www.radix-ui.com/docs/primitives/components/tabs

The new component is, for now `NewTabs`.
Since the sub-component don't clash with the older `TabItem` I kept the original names so when we finished the replacement we can only rename `NewTabs` in `Tabs`

Since radix also requires a tab `value` that is used as an index to link the tabs button with their content, I've used this `value` to add some classes to both the buttons and the content too. 

## Checklist

## Steps to Test

Example:

1. Pull down PR.
1. `npm run dev`.
1. Open Storybook on the `NewTabs` component https://deploy-preview-130--vip-design-system-components.netlify.app/?path=/story/newtabs--default
1. Compare the visual style of the tabs [with the old one](https://deploy-preview-130--vip-design-system-components.netlify.app/?path=/story/tabs--default)
1. Verify that the tabs are accessible, that they are announced and you can move with the left/right arrow when focus is on them

### DIFF
--- a/src/system/NewTabs/Tabs.js
+++ b/src/system/NewTabs/Tabs.js
@@ -1,0 +1,29 @@
+/** @jsxImportSource theme-ui */
+
+/**
+ * External dependencies
+ */
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+import * as TabsPrimitive from '@radix-ui/react-tabs';
+
+/**
+ * Internal dependencies
+ */
+
+const NewTabs = ( { defaultValue = null, className = null, sx, ...props } ) => (
+	<TabsPrimitive.Root
+		defaultValue={ defaultValue }
+		className={ classNames( 'vip-tabs-component', className ) }
+		{ ...props }
+	/>
+);
+
+NewTabs.propTypes = {
+	className: PropTypes.any,
+	sx: PropTypes.object,
+	variant: PropTypes.string,
+	defaultValue: PropTypes.string,
+};
+
+export { NewTabs };
--- a/src/system/NewTabs/Tabs.js
+++ b/src/system/NewTabs/Tabs.js
@@ -13,14 +13,20 @@ import React from 'react';
  */
 
 const NewTabs = React.forwardRef(
-	( { defaultValue = null, className = null, sx, ...props }, ref ) => {
+	(
+		{ children, onValueChange = undefined, defaultValue = undefined, className = null, sx = {} },
+		ref
+	) => {
 		return (
 			<TabsPrimitive.Root
 				ref={ ref }
 				defaultValue={ defaultValue }
+				onValueChange={ onValueChange }
 				className={ classNames( 'vip-tabs-component', className ) }
-				{ ...props }
-			/>
+				sx={ { ...sx } }
+			>
+				{ children }
+			</TabsPrimitive.Root>
 		);
 	}
 );
@@ -28,7 +34,9 @@ const NewTabs = React.forwardRef(
 NewTabs.propTypes = {
 	className: PropTypes.any,
 	sx: PropTypes.object,
-	defaultValue: PropTypes.string,
+	defaultValue: PropTypes.node,
+	onValueChange: PropTypes.func,
+	children: PropTypes.node,
 };
 
 NewTabs.displayName = 'NewTabs';
--- a/src/system/NewTabs/Tabs.js
+++ b/src/system/NewTabs/Tabs.js
@@ -22,7 +22,6 @@ const NewTabs = ( { defaultValue = null, className = null, sx, ...props } ) => (
 NewTabs.propTypes = {
 	className: PropTypes.any,
 	sx: PropTypes.object,
-	variant: PropTypes.string,
 	defaultValue: PropTypes.string,
 };
 
--- a/src/system/NewTabs/Tabs.js
+++ b/src/system/NewTabs/Tabs.js
@@ -6,17 +6,23 @@
 import classNames from 'classnames';
 import PropTypes from 'prop-types';
 import * as TabsPrimitive from '@radix-ui/react-tabs';
+import React from 'react';
 
 /**
  * Internal dependencies
  */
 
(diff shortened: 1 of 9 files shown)


src/system/NewTabs/TabsTrigger.js
```diff
@@ -0,0 +1,50 @@
...
+			pb: 3,
+			border: 'none',
+			color: 'muted',
+			transform: 'translateY(1px)',
+			'&[data-state="active"]': {
+				color: 'heading',
+				borderBottom: '1px solid',
+				borderColor: 'link',
+			},
+			':hover': { fontWeight: 'body', color: 'heading' },
```
djalmaaraujo:
```suggestion
			':hover': { fontWeight: 'body', color: 'heading' },
			'&:focus': theme => theme.outline,
			'&:focus-visible': theme => theme.outline,
```

src/system/NewTabs/TabsTrigger.js
```diff
@@ -0,0 +1,50 @@
...
+			fontSize: 2,
+			px: 0,
+			pb: 3,
+			border: 'none',
+			color: 'muted',
+			transform: 'translateY(1px)',
+			'&[data-state="active"]': {
+				color: 'heading',
+				borderBottom: '1px solid',
+				borderColor: 'link',
```
djalmaaraujo: I think you can leave the focus do this job for you, you can remove the border

src/system/NewTabs/Tabs.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+import * as TabsPrimitive from '@radix-ui/react-tabs';
+
+/**
+ * Internal dependencies
+ */
+
+const NewTabs = ( { defaultValue = null, className = null, sx, ...props } ) => (
```
djalmaaraujo: It would be nice to add the `forwardRef` capability, and pass it to the first container.

Example:

```
const Card = React.forwardRef( ( { variant = 'primary', sx = {}, className, ...props }, ref ) => {
	return (
		<Box
			ref={ ref }
```

src/system/NewTabs/Tabs.stories.jsx
```diff
@@ -0,0 +1,28 @@
...
+import { NewTabs, TabsTrigger, TabsList, TabsContent, Text } from '..';
+
+export default {
+	title: 'NewTabs',
+	component: NewTabs,
+};
+
+export const Default = () => (
+	<NewTabs defaultValue="all">
+		<TabsList aria-label="See all the content">
```
djalmaaraujo: Can you make the `aria-label` or a `title` attribute required for this? People might forget to add this attribute. Title is nice, or caption.

djalmaaraujo: It's recommended always to have a label for the tabs.

src/system/NewTabs/Tabs.js
djalmaaraujo: Currently, we are avoiding having `...props`, since this creates confusion about the components' API. If we don't need all radix API, we can use only what is needed?

src/system/NewTabs/Tabs.stories.jsx
```diff
@@ -0,0 +1,28 @@
+/**
```
djalmaaraujo, line 1: I saw radix supports orientation vertical, do you think it's possible to have or we can do in a future version?

djalmaaraujo (changes requested):
@andrea-sdl AWESOME Work!

I left some comments/recommendations.

src/system/NewTabs/Tabs.stories.jsx
andrea-sdl: @djalmaaraujo do we already have a design example for the vertical tabs? I tried playing around and we can support them but I feel that if we have a design to follow we'll get a better output.

src/system/NewTabs/Tabs.stories.jsx
andrea-sdl: added in https://github.com/Automattic/vip-design-system/pull/130/commits/c441b2df91e7d33733a8a0eba6f19fcad8fe5036 although I wasn't able to remove the `...props` entirely (I wonder why it stops working if I do `title={ title }` )

src/system/NewTabs/TabsTrigger.js
andrea-sdl: I tried removing the `borderBottom` once I added the focus part but I wonder if this is the expected behavior (having the blue box)
Screenshot 2022-11-08 at 18 44 56

djalmaaraujo: @andrea-sdl Having the focus outline is the expected behavior, yes.

src/system/NewTabs/Tabs.stories.jsx
djalmaaraujo: @andrea-sdl I don't think so, so feel free to ignore for now

src/system/NewTabs/TabsTrigger.js
andrea-sdl: @djalmaaraujo not removing the border since it allows to see which tabs is active when focus is not present.
I changed the code to follow the example from radix (using `box-shadow` instead of border, very nice and doesn't need `transform`)

src/system/NewTabs/Tabs.stories.jsx
andrea-sdl: @djalmaaraujo fixed this too (it was an issue with the `children` not being passed. We now have the title attribute to use.

djalmaaraujo (approved):
This looks amazing. Awesome new component!
Thank you @andrea-sdl 

Let me know if you need help to publish a new version.

PR #142, opened 2022-11-23
Revamp Radio Component with A11Y

djalmaaraujo:
## Description

image

We are revamping the `Radio` component. Radio components only make sense to be used with two or more. That's the nature of Radios; you need two or more options.

The current implementation lets you choose only one Radio button. In this new implementation, you must pass an array of options (similar to other components). This will provide more clarity and avoid forms with one radio button, and this new implementation is accessible.

**BREAKING CHANGE**: The new implementation of the `<Radio>` will break current usage. Once you upgrade this component, you need to change the implementation. Follow the Storybook story to understand the new API.

## Checklist

## Steps to Test

1. Pull down PR.
2. `npm run dev`.
3. Open http://localhost:6006/?path=/story/form-radio--default
4. Test the functionality. Make sure you can click/select using the mouse or keyboard the toggle and the label

### DIFF
--- a/src/system/Form/Label.js
+++ b/src/system/Form/Label.js
@@ -8,15 +8,20 @@ import React from 'react';
 /**
  * Internal dependencies
  */
-import { Heading } from '..';
 
-const Label = React.forwardRef( ( props, forwardRef ) => (
-	<Heading
-		variant="h4"
-		as="label"
-		sx={ { display: 'block', mb: 2, color: 'muted' } }
+const Label = React.forwardRef( ( { sx, ...rest }, forwardRef ) => (
+	<label
+		sx={ {
+			fontWeight: 500,
+			fontSize: 2,
+			lineHeight: 1.5,
+			display: 'block',
+			mb: 2,
+			color: 'muted',
+			...sx,
+		} }
 		ref={ forwardRef }
-		{ ...props }
+		{ ...rest }
 	/>
 ) );
 
--- a/src/system/Form/Label.js
+++ b/src/system/Form/Label.js
@@ -4,6 +4,7 @@
  * External dependencies
  */
 import React from 'react';
+import PropTypes from 'prop-types';
 
 /**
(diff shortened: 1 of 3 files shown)


simonwheatley:
Thanks for the classes on each item, @djalmaaraujo

src/system/Form/Radio.js
```diff
@@ -1,25 +1,160 @@
...
+	children: PropTypes.shape( { props: { className: PropTypes.any, sx: PropTypes.object } } )
+		.isRequired,
+};
+
+const Radio = React.forwardRef(
+	(
+		{ disabled, defaultValue, onChange, name = '', options = [], className, ...props },
+		forwardRef
+	) => {
+		const renderedOptions = useMemo(
```
brunobasto: @djalmaaraujo I wonder if the overhead of `useMemo` is beneficial here. I would guess the options array will likely be small (< 5), and I don't see any heavy processing happening inside the loop.

src/system/Form/Label.js
```diff
@@ -4,22 +4,32 @@
...
  */
-import { Heading } from '..';
 
-const Label = React.forwardRef( ( props, forwardRef ) => (
-	<Heading
-		variant="h4"
-		as="label"
-		sx={ { display: 'block', mb: 2, color: 'muted' } }
+const Label = React.forwardRef( ( { sx, ...rest }, forwardRef ) => (
+	<label
```
brunobasto, line 14: excellent!

brunobasto (approved):
Looks amazing, @djalmaaraujo! Left a minor comment.

src/system/Form/Radio.js
djalmaaraujo: @brunobasto Might not be that much, indeed. The main reason I added it is because, usually, people do `onChange` with an inline-function. The parent will make this component re-render often, mainly because it's a form with many interactions.

PR #150, opened 2022-12-01
Fix the input placeholder color for a better contrast on light theme

djalmaaraujo:
## Description

Our current placeholder color has a lower contrast than 4.5 when using the Input background as white. This PR fixes the placeholder color with #757575, which is at least 4.6.

The design team might adjust this, but doing this now can fix a lot of issues for accessibility.

image

## Steps to Test

1. Pull down PR.
2. `npm run dev`.
3. Open http://localhost:6006/?path=/story/form-input--default
4. Inspect the Input CSS. The computed color for the placeholder should match #757575

image

### DIFF
--- a/src/system/Form/Input.js
+++ b/src/system/Form/Input.js
@@ -33,6 +33,7 @@ const inputStyles = {
 	},
 	'&::placeholder': {
 		color: 'placeholder',
+		opacity: 1,
 	},
 };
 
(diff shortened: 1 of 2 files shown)


luiztiago (approved):

rianrietveld:
@djalmaaraujo 
Approved :-)

PR #160, opened 2022-12-15
[PIE-3491] Button aria-disabled

luiztiago:
## Description

https://vipjira.atlassian.net/browse/PIE-3491?atlOrigin=eyJpIjoiMTQ1OTRiYjNlNTI0NDc5Nzk5NGYyYjNmNjAxZjg4NDMiLCJwIjoiaiJ9

- I also added the controls to the Button storybook. Now, we can control the props to test them.

### References

https://a11y-101.com/development/aria-disabled
https://www.smashingmagazine.com/2021/08/frustrating-design-patterns-disabled-buttons/
https://css-tricks.com/making-disabled-buttons-more-inclusive/

## Checklist

## Steps to Test

1. Pull down PR.
1. `npm run dev`.
1. Open Storybook > https://deploy-preview-160--vip-design-system-components.netlify.app/?path=/story/button--default&args=disabled:true
1. Open Controls tab and set "disabled" as true. Click on the button and you should not be redirected to google
1. Set "disabled" as false. Click on the button and you should redirected to google

### DIFF
--- a/src/system/Button/Button.js
+++ b/src/system/Button/Button.js
@@ -12,9 +12,11 @@ const Button = React.forwardRef( ( { disabled, onClick, sx, ...props }, forwardR
 	const handleOnClick = useCallback(
 		event => {
 			if ( disabled ) {
-				event.preventDefault();
-			} else if ( onClick ) {
-				onClick( event );
+				return event.preventDefault();
+			}
+
+			if ( onClick ) {
+				return onClick( event );
 			}
 		},
 		[ disabled, onClick ]
--- a/src/system/Button/Button.js
+++ b/src/system/Button/Button.js
@@ -32,7 +32,7 @@ const Button = React.forwardRef( ( { disabled, sx, ...props }, forwardRef ) => {
 				},
 				'&:focus': theme => theme.outline,
 				'&:focus-visible': theme => theme.outline,
-				'&:disabled': {
+				'&[aria-disabled]': {
 					opacity: 0.7,
 					cursor: 'not-allowed',
 					pointerEvents: 'all',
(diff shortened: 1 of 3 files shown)


src/system/Button/Button.test.js
```diff
@@ -18,4 +18,16 @@ describe( '<Button />', () => {
 		// Check for accessibility issues
 		await expect( await axe( container ) ).toHaveNoViolations();
 	} );
+
+	it( 'renders the Button with aria-disabled prop', async () => {
```
djalmaaraujo:
```suggestion
	it( 'renders the Button with disabled prop', async () => {
```

src/system/Button/Button.js
```diff
@@ -3,44 +3,60 @@
...
+	const handleOnClick = useCallback(
+		event => {
+			if ( disabled ) {
+				event.preventDefault();
+			} else if ( onClick ) {
+				onClick( event );
+			}
+		},
+		[ disabled, onClick ]
+	);
```
djalmaaraujo:
```suggestion
	const handleOnClick = useCallback(
		event => {
			if ( disabled ) {
			  return event.preventDefault();
			}
			
			if ( onClick ) {
			  return onClick( event );
			}
		},
		[ disabled, onClick ]
	);
```

Could this be simpler? I think if it's not disabled, you still want to call the `onClick`, right? So, in this case could not be a `else if`.

djalmaaraujo (changes requested):
Thank you for working on this.
Left some minor changes.

djalmaaraujo (approved):

PR #164, opened 2023-01-10
Update/valet theme

djalmaaraujo:
## Description

Valet theme updates:

- Input default text token added to input.text.default.
- Input default text color darkened to match text.primary (a small additional increase in font weight may be required here if the placeholder and the input default text still resembel one another). 
- Input label token added to input.label.
- Muted inverse text token added to text.inverse-secondary.
- Corrected contrast issues on Info notice's link color.

## Steps to Test

1. Notice has no color contrast issues anymore: http://localhost:6006/?path=/story/notice--default


luiztiago (approved):
LGTM, great job!

PR #169, opened 2023-01-23
Fix spinner color

djalmaaraujo:
## Description

We need to keep the Spinner as the link color. Looks like the link color is the most efficient for all situations.

## Checklist

## Steps to Test

1. Pull down PR.
2. `npm run dev`.
3. Open the [Spinner](http://localhost:6006/?path=/story/spinner--default) story. it should be orange as the link color


kat3samsin (approved):
Works! Thanks for fixing so fast!

PR #171, opened 2023-01-25
Show the selected value in the input of Form.Autocomplete 

andrea-sdl:
## Description

Right now if you start writing in the select and the focus away from it, the old value is still selected and you have no chance of knowing that. 
With this PR I try to introduce a new behaviour where if the select loses focus and no selection has been made, then the select is resetted. 

This might not be the best solution, but it will allow us to fix some confusion while using the autocomplete.

I plan to also fix the original component and remove this. 

KNOWN BUG: If you input some data and you use the keyboard to move and make a selection, the input field will be resetted to either the selected value or empty if no selection  has been  made. 

### Before
Before Screenshot screencast 2023-01-25 15-43-55

### After

After Screenshot screencast 2023-01-25 16-09-58

**This PR also introduces support for controls in Storybook for the Autocomplete component.**
This way you can test the `resetOnBlur` filter for every example (the only one not supported at the moment is the Debounce)
CleanShot 2023-02-10 at 18 45 18@2x

## Checklist

## Steps to Test

1. Pull down PR.
1. `npm run dev`.
1. Open Storybook for the component http://localhost:6006/?path=/story/form-autocomplete--default 
1. Try to select a field, then edit the input and move the focus away (clicking in the page or moving away with a tab). The field should reset to the selected state
1.After selecting a state, delete all the contents, the when losing focus it should keep the empty contents.


luiztiago (approved):
Excellent, great job! :shipit:

luiztiago:
@andrea-sdl I agree with your comment. I think it should be better to fork the main component and work on some code improvements. The accessibility side of this component is awesome, but the code 🤦

kat3samsin:
This is great @andrea-sdl! We can now handle empty values!

djalmaaraujo:
@andrea-sdl The only comment I have is if we should set the value as `''` or null, because the options are objects, not simple strings.

cc: @luiztiago

src/system/NewForm/FormAutocomplete.js
```diff
@@ -153,29 +160,44 @@ const FormAutocomplete = React.forwardRef(
...
 			inputValue => {
 				if ( inputValue ) {
+					setSelectedValue( inputValue );
+					setInputQuery( inputValue );
+					onChange( getOptionByLabel( inputValue ), inputValue );
+					setIsDirty( false );
+				} else if ( resetOnNoMatch && inputQuery !== selectedValue ) {
+					// reset the content if there's no match
+					setSelectedValue( '' );
+					setInputQuery( '' );
```
andrea-sdl: > The only comment I have is if we should set the value as '' or null, because the options are objects, not simple strings.
@djalmaaraujo you mean here? Keep in mind that selectedValue/InputQuery are string or simple types. The object is the one we pass on the `onChange`. Which is retrieved through the `getOptionByLabel`

andrea-sdl: @djalmaaraujo with the latest changes I'm sending `null` as you proposed.

kat3samsin (approved):
Works! Thanks for working on this!

GaryJones:
Going only by the After image, if I type in `Strawberry Chocolate Vanila` (with the typo), and then select away, does that mean all my typing has been lost?

andrea-sdl:
> Going only by the After image, if I type in `Strawberry Chocolate Vanila` (with the typo), and then select away, does that mean all my typing has been lost?

@GaryJones Yes. This partly happened also with the original component ([demos](https://alphagov.github.io/accessible-autocomplete/examples/))
CleanShot 2023-02-08 at 15 02 17
From the video is not clear but I didn't click on the element, I moved away from it with the keyboard by pressing TAB.

The main difference here is that we reset to the selected value too.
BTW: I agree this is still far from perfect, on the other hand, right now, we have no way to be sure of what's selected.

src/system/NewForm/FormAutocomplete.js
```diff
@@ -239,6 +288,11 @@ const FormAutocomplete = React.forwardRef(
 			);
 		}, [] );
 
+		useEffect( () => {
+			global.document.querySelector( `#${ forLabel }` ).addEventListener( 'blur', () => {
+				resetInputState();
```
kat3samsin, line 294: Thanks for this I also need this in the multiselect!

kat3samsin (changes requested):
@andrea-sdl Awesome work! ~~Just a minor comment on the source parameter.~~ Nevermind this comment. 😛  🎉

kat3samsin (approved):

djalmaaraujo:
@andrea-sdl I am not sure why @dabowman commits are showing in this PR, did you used `develop` as the source branch of this PR? To avoid any confusion, please, recreate from trunk. We don't use develop in this repo.

David uses this for Valet theme.

cc: @dabowman Another reason we might need to move this push figma tokens from this repo, might break things ;)

andrea-sdl:
@djalmaaraujo that was an error on my side done in a3cd373c52a34be609feb71f085fb01f42161716 and reverted in f7c152a62d801c24ab7a78ad8d69862cdce174eb 

I confused `develop` with `trunk` :( 

Git will still show the history even though no changes are being applied.
One way around this is to rewrite it entirely (but it will take a big rebase) or create a clean PR.

PR #175, opened 2023-01-27
Add support for Wizard title autofocus when switching tabs

andrea-sdl:
## Description
This PR Adds support for autofocusing the Wizard step title without the need to pass down many ref.

Useful because when you navigate with the keyboard you want the steps to be focused.

## Checklist

## Steps to Test

Example:

1. Pull down PR.
1. `npm run dev`.
1. Open Storybook. http://localhost:6006/?path=/story/wizard--with-title-auto-focus
1. Navigate between the steps with keyboard only, each time you move back and forth if you press tab you should be focusing the button (because it's the next component after the title)
1. Always with the keyboard, disable the autofocus in the example, if you play a couple of times (going back and forth between the continue/back buttons) you'll notice that at one point the focus goes away from the wizard and focuses on the Autofocus selectbox

### DIFF
--- a/src/system/Wizard/Wizard.js
+++ b/src/system/Wizard/Wizard.js
@@ -3,7 +3,7 @@
 /**
  * External dependencies
  */
-import React, { useLayoutEffect } from 'react';
+import React, { useLayoutEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { MdArrowForward } from 'react-icons/md';
@@ -26,11 +26,12 @@ const Wizard = React.forwardRef(
 		},
 		forwardRef
 	) => {
-		const [ initialStep, setInitialStep ] = React.useState( activeStep );
+		const didMount = useRef( false );
+		// didMount helps us to track the initial render, so we can focus the title only subsequent renders
+		// to avoid stealing the focus from the page we're in.
 		useLayoutEffect( () => {
-			if ( titleAutofocus && activeStep !== initialStep ) {
-				// After the initial page load, the initial step can be focused
-				setInitialStep( -1 );
+			if ( ! didMount.current ) {
+				didMount.current = true;
 			}
 		}, [ activeStep ] );
 		return (
@@ -70,9 +71,7 @@ const Wizard = React.forwardRef(
 							subTitle={ subTitle }
 							title={ title }
 							titleVariant={ titleVariant }
-							shouldFocusTitle={
-								titleAutofocus && activeStep !== initialStep && index === activeStep
-							}
+							shouldFocusTitle={ titleAutofocus && didMount.current }
 						>
 							{ children }
 						</WizardStep>
--- a/src/system/Wizard/Wizard.js
+++ b/src/system/Wizard/Wizard.js
@@ -3,7 +3,7 @@
 /**
  * External dependencies
  */
-import React from 'react';
+import React, { useLayoutEffect } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { MdArrowForward } from 'react-icons/md';
@@ -14,7 +14,25 @@ import { MdArrowForward } from 'react-icons/md';
 import { Box, WizardStep, Flex, WizardStepHorizontal } from '..';
 
 const Wizard = React.forwardRef(
-	( { steps, activeStep, variant, completed = [], className = null, ...props }, forwardRef ) => {
+	(
+		{
+			steps,
+			activeStep,
+			variant,
+			completed = [],
+			className = null,
+			titleAutofocus = false,
+			...props
+		},
+		forwardRef
+	) => {
+		const [ initialStep, setInitialStep ] = React.useState( activeStep );
+		useLayoutEffect( () => {
+			if ( titleAutofocus && activeStep !== initialStep ) {
+				// After the initial page load, the initial step can be focused
+				setInitialStep( -1 );
+			}
+		}, [ activeStep ] );
 		return (
 			<Box className={ classNames( 'vip-wizard-component', className ) } ref={ forwardRef }>
 				{ variant === 'horizontal' ? (
@@ -52,6 +70,7 @@ const Wizard = React.forwardRef(
 							subTitle={ subTitle }
 							title={ title }
 							titleVariant={ titleVariant }
+							shouldFocusTitle={ titleAutofocus && activeStep !== initialStep }
 						>
 							{ children }
 						</WizardStep>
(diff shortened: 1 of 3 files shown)


src/system/Wizard/WizardStep.js
```diff
@@ -67,6 +81,7 @@ const WizardStep = React.forwardRef(
 							fontSize: 2,
 							fontWeight: active ? 'bold' : 'normal',
 						} }
+						ref={ titleRef }
```
kat3samsin, line 85: I think you forgot to add this to make the element focusable.
`tabIndex={ -1 }`

src/system/Wizard/WizardStep.js
```diff
@@ -32,7 +42,11 @@ const WizardStep = React.forwardRef(
 		} else if ( active ) {
 			color = 'heading';
 		}
-
+		useLayoutEffect( () => {
+			if ( active && titleRef?.current && shouldFocusTitle ) {
```
kat3samsin, line 47: I think checking of `active` is enough and we can remove `shouldFocusTitle` param.

kat3samsin (changes requested):
Thank you so much for working on this @andrea-sdl! 🎉
I have a few suggestions I added in the code.

src/system/Wizard/WizardStep.js
andrea-sdl: @kat3samsin It would be active on the first load too, though, right? In this case we will forcefully focus on the first step when the user has yet to engage with the page. 

I would expect the user to prefer to navigate/read the page and _then_ visiting the first step. 

How would you handle this if you remove the shouldFocusTitle?

kat3samsin: I see what you mean, we should focus on the Wizard Header first rather than on the Wizard Step title on load.

src/system/Wizard/Wizard.js
```diff
@@ -52,6 +70,7 @@ const Wizard = React.forwardRef(
 							subTitle={ subTitle }
 							title={ title }
 							titleVariant={ titleVariant }
+							shouldFocusTitle={ titleAutofocus && activeStep !== initialStep }
```
djalmaaraujo: What is the reason for this parameter called `titleAutofocus ` and the other in the other file is `shouldFocusTitle`?

djalmaaraujo (commented):
@andrea-sdl Thanks for this suggestion.

Can you take a look at this post and check if we can do something similar? https://www.ombulabs.com/blog/react/javascript/how-to-export-instance-methods-from-react-function-components.html

Exposing a method would be better. I am concerned about adding `tabIndex` to all Wizard on the page; I think we had this before, and it was causing A11Y issues.

Another option is to control from the outside of the component, in a VIP Dashboard hook. This way, we can keep the `focusOnRender` or `titleAutoFocus` as you created and use the hook to control when to focus.

I am not sure to be honest, I will investigate a little more.

src/system/Wizard/WizardStep.js
djalmaaraujo: I belive we should add the tabIndex only when the autoFocus is ON, because this is not good for A11Y having -1 across the code.

src/system/Wizard/Wizard.js
andrea-sdl: Good question! 
`titleAutofocus` is the feature of the component that can be on and off.
`shouldFocusTitle` indicates if/when a single step should focus its title.

Have a better idea for naming? I'm all ears

src/system/Wizard/WizardStep.js
andrea-sdl: @djalmaaraujo @kat3samsin added the optional TabIndex in https://github.com/Automattic/vip-design-system/pull/175/commits/e39c3946c2a452ac7213d510fae1217be3855d70

Thanks!

src/system/Wizard/Wizard.js
```diff
@@ -14,7 +14,25 @@ import { MdArrowForward } from 'react-icons/md';
...
+			completed = [],
+			className = null,
+			titleAutofocus = false,
+			...props
+		},
+		forwardRef
+	) => {
+		const [ initialStep, setInitialStep ] = React.useState( activeStep );
+		useLayoutEffect( () => {
+			if ( titleAutofocus && activeStep !== initialStep ) {
```
kat3samsin: I wonder why we don't want to focus on the initial step. I find it's better to focus on the next step to direct the users. Right now, when I click on the wizard option, I focus on the navigation sidebar.

kat3samsin: I forgot to submit this comment a few days ago, apologies.

andrea-sdl: @kat3samsin Maybe I should rename it. So, this is the initial status of the wizard. The idea is that, when the wizard is first shown, we shouldn't focus on the title. This is because otherwise we might grab the focus from other elements in the page. 
We want the focus to be controlled when moving inside the wizard. Or at least that's how I envisioned.

kat3samsin: Got it, I see what you mean. I wonder, though if we should direct users to the current step if for example they are in the middle of launching [site](https://dashboard.wpvip.com/apps/6618/production/network-sites/launch/5), I needed to navigate to the previous steps before I could go to the current step. What do you think?

andrea-sdl: I now understand what you  meant. The issue is that I use activestep as the starting point where I should have some firstLoad state. Will change that logic so that it works if we start the wizard on a different step

andrea-sdl: > I needed to navigate to the previous steps before I could go to the current step. What do you think?

I think this is accessible in my view, but I guess this is debatable.
I normally scan the other headers to have an understanding of the status, that's why I don't think it's a bad idea to -not- focus on first render. 

An alternative to this would be to make the `titleAutofocus` be an enum (or not a boolean) so that we can require different behaviours. 

In any case, if you pass the title not as a string, but as a component the behaviour can be entirely customized if needed.
The `titleAutofocus` kicks in only to make the "default" more accessible.

Inspired by the comment I also removed the dependency on `activeStep` to track the first render and used a more generic approach here https://github.com/Automattic/vip-design-system/pull/175/commits/c25680c49f8289f80fc3657db39493fb8f311481

kat3samsin: > I think this is accessible in my view, but I guess this is debatable.

Yes, I agree this depends on the use case of the wizard.

>Inspired by the comment I also removed the dependency on activeStep to track the first render and used a more generic approach here https://github.com/Automattic/vip-design-system/commit/c25680c49f8289f80fc3657db39493fb8f311481

Thanks for updating!

kat3samsin (approved):
🚢

PR #179, opened 2023-02-07
Autocomplete with Multiselect

kat3samsin:
## Description

**Autocomplete with Multiselect**

The new component `Form.AutocompleteMulti` is based on Autocomplete from Gov.UK to support components in VIP Dashboard that require multi-select (Custom Install Certificate > Activate Domains).

I need your feedback on how to make this better. 🙏🏻

image
image
image

Features:

- Removes item on the list if already selected.
- Displays the current # of items selected.
- Supports `CommonFormNoticeErrors` and `Validation` (hint)

## Checklist

## Steps to Test

1. Pull down PR.
2. `npm run storybook`

OR

1. Navigate to [https://deploy-preview-179--vip-design-system-components.netlify.app/?path=/story/form-autocompletemulti--with-dynamic-data](https://deploy-preview-179--vip-design-system-components.netlify.app/?path=/story/form-autocompletemulti--with-dynamic-data)


donaghokeeffe:
SCR-20230209-ddf
image
About input's default state & placeholder text/icon.

donaghokeeffe:
SCR-20230209-dci
About the "Selected value" element in the demo.

donaghokeeffe:
SCR-20230209-dfm
About a "# of selected" indicator.

donaghokeeffe:
SCR-20230209-di0
About the display of the "selected item tags". Is a "x" remove icon button accessible in this case? (as seen in the zendesk implementation).

donaghokeeffe:
SCR-20230209-daj
I think you've mentioned this in the PR description above as something you're aware of. Just thought I'd post it anyway.

src/system/NewForm/FormAutocompleteMultiselect.stories.jsx
```diff
@@ -0,0 +1,107 @@
...
+	{ value: 'pistachio', label: 'Pistachio' },
+	{ value: 'bubblegum', label: 'Bubblegum' },
+	{ value: 'ube', label: 'Ube' },
+	{ value: 'mango', label: 'Mango' },
+	{ value: 'buko', label: 'Buko' },
+	{ value: 'durian', label: 'Durian' },
+	{ value: 'lecheflan', label: 'Leche Flan' },
+	{
+		value: 'averylongicecreamflavor',
+		label: 'This is a very long ice cream flavorrrr with no name but it tastes goooooooood...',
```
andrea-sdl: @kat3samsin if this will be used for domains we should also add an example with no spaces. Spaces might help with wrapping, depending on how the `whitespace` CSS property is configured, whereas with no space we hardly have wrapping.

src/system/NewForm/FormAutocompleteMultiselect.stories.jsx
```diff
@@ -0,0 +1,107 @@
...
+
+	return (
+		<>
+			<Form.Root>
+				<div sx={ { width } }>
+					<Form.AutocompleteMulti
+						forLabel="form-autocompletemultiselect"
+						label={ label }
+						onChange={ obj => {
+							setSelectedValue( obj );
```
andrea-sdl: @kat3samsin what do you think about changing this to `setSelectedValues`? This way it's clear that we are passing an array in the example.

src/system/NewForm/FormAutocompleteMultiselect.stories.jsx
```diff
@@ -0,0 +1,107 @@
...
+	const [ selectedValue, setSelectedValue ] = useState( [] );
+
+	return (
+		<>
+			<Form.Root>
+				<div sx={ { width } }>
+					<Form.AutocompleteMulti
+						forLabel="form-autocompletemultiselect"
+						label={ label }
+						onChange={ obj => {
```
andrea-sdl, line 68: Do we have a way to know what was the action when it was triggered? I can see it being useful to know if it was an addition or a deletion and which object was added/deleted

kat3samsin:
> SCR-20230209-dci About the "Selected value" element in the demo.

This is not part of the component. It's only for the storybook to make sure that the values selected are correct.

src/system/NewForm/FormAutocompleteMultiselect.stories.jsx
kat3samsin: Thanks for the suggestion! I'll create an example with long domain names.

src/system/NewForm/FormAutocompleteMultiselect.stories.jsx
kat3samsin: You're right it's much clearer. I'll update!

kat3samsin:
Updates:
- Added count of selected items.
- Cleared input on option select.
- Adding filtering of items, does not show if already selected.

image

kat3samsin:
@djalmaaraujo comments

- [x] When you remove an item from the list, ensure the focus is returned to something that makes sense (not sure what would be); Maybe we should move the focus to the list with the item chosen selected, so it's clear that he pick that one. But also, not sure if the best UX, since he needs to TAB back to select more.
_I find that most autocomplete does not remove the focus to the input like in zendesk, to be able to easily add more items._

- [x] I guess you did not reset the state yet because Andreas PR was not merged

- [x] The selected options should be a button, not a link. Links takes users to somewhere, like a route

- [x] I am not sure how useful the <badge> at the top can me used for blind users.

- [x] Maybe when searching, we should omit already selected items from the list

- [ ] When we select, there's no indication that the item was selected, or anything about the list below

donaghokeeffe:
image
I know it's a bit long, but the existing text instruction feels a bit redundant.

donaghokeeffe:
image
I think it should only reset to show the placeholder once I've actually selected an item. Until then the field should show either the content I've typed (like it does when using the mouse to move over the results list), or it should change from what I've typed into what I've currently selected (the value of the row in black).

donaghokeeffe:
image
To reproduce, just type "ch" and the field will bring up the shown results above. Then delete the text and tab away from the field. If you tab back to the field again that list of results appears before you type anything, which can be confusing.

kat3samsin:
> I think it should only reset to show the placeholder once I've actually selected an item. Until then the field should show either the content I've typed (like it does when using the mouse to move over the results list), or it should change from what I've typed into what I've currently selected (the value of the row in black).

Thank you for the feedback @donaghokeeffe! I didn't realize this behavior, I'll try to fix it.

kat3samsin:
> image I know it's a bit long, but the existing text instruction feels a bit redundant.

This is the default message but can be adjusted on the consumer side.

kat3samsin:
> To reproduce, just type "ch" and the field will bring up the shown results above. Then delete the text and tab away from the field. If you tab back to the field again that list of results appears before you type anything, which can be confusing.

I appreciate the feedback! I'll find a way to fix the behavior.

src/system/NewForm/FormAutocompleteMultiselect.stories.jsx
kat3samsin: Makes sense, I see the usefulness of this feature. Thank you for pointing this out! I'll try to add tracking of selectedOptions (increase/decrease).

Caitboard:
I can tell a lot of thought has been put into this, which is great, well done! I have a few notes that need improvement:

- [x] The color contrast of the placeholder is currently too low (#A09E9E vs #FFF background), which results in 2.7:1 contrast. This needs to be at least 4.5:1
- [ ] When testing with VoiceOver there are some really great clues as to how this element works. However, something is causing the screen reader texts / ARIA hints to be read multiple times. When there are two results available this is doable, but when I have 6 results available VoiceOver starts getting stuck, which could cause the user to select the wrong item. It takes forever for it to be done reading everything, even though I have already tabbed to the next button. I have added a screen recording to showcase what is happening in Slack. [Watch screen recording in Slack](https://level-level.slack.com/archives/C036D28RW5S/p1676886316374499). If you can get the info to be read only once I think this could work great.

kat3samsin:
>When testing with VoiceOver there are some really great clues as to how this element works. However, something is causing the screen reader texts / ARIA hints to be read multiple times. When there are two results available this is doable, but when I have 6 results available VoiceOver starts getting stuck, which could cause the user to select the wrong item. It takes forever for it to be done reading everything, even though I have already tabbed to the next button. I have added a screen recording to showcase what is happening in Slack. [Watch screen recording in Slack](https://level-level.slack.com/archives/C036D28RW5S/p1676886316374499). If you can get the info to be read only once I think this could work great.

@djalmaaraujo and I paired with this issue, and we discovered that this is due to testing the component using Storybook which uses an iframe. You can find more details of the issue [here](https://github.com/phetsims/a11y-research/issues/132).

@Caitboard  Could you test it again outside of the iframe? [http://localhost:6006/iframe.html?args=&id=form-autocompletemulti--default&viewMode=story](http://localhost:6006/iframe.html?args=&id=form-autocompletemulti--default&viewMode=story).

src/system/NewForm/FormAutocompleteMultiselect.js
```diff
@@ -0,0 +1,393 @@
...
+		useEffect( () => {
+			onChange(
+				selectedOptions,
+				selectedOptions.map( option => option?.label || option )
+			);
+			resetInputState();
+		}, [ selectedOptions ] );
+
+		return (
+			<div className={ classNames( 'vip-form-autocomplete-component', className ) }>
```
djalmaaraujo, line 386: I believe we need to update this to `vip-form-autocomplete-multi-component`.

kat3samsin:
Added VO on item add `item added to list`.
Fix focus on item removal.

djalmaaraujo (approved):

Caitboard:
> @djalmaaraujo and I paired with this issue, and we discovered that this is due to testing the component using Storybook which uses an iframe. You can find more details of the issue [here](https://github.com/phetsims/a11y-research/issues/132).
> 
> @Caitboard Could you test it again outside of the iframe? http://localhost:6006/iframe.html?args=&id=form-autocompletemulti--default&viewMode=story.

Argh, that is annoying! Glad you figured it out. It works perfectly. I would remove just this one sentence from the filter button: "Press Shift Tab to add more". This is already clear for screen reader users as they navigate through the elements. We want to avoid an information overload, so it's best to avoid this. Besides that, this solution is great. Amazing work!

image

kat3samsin:
> Argh, that is annoying! Glad you figured it out. It works perfectly. I would remove just this one sentence from the filter button: "Press Shift Tab to add more". This is already clear for screen reader users as they navigate through the elements. We want to avoid an information overload, so it's best to avoid this. Besides that, this solution is great. Amazing work!
> 
> image

Thank you for the feedback @Caitboard we'll update in a new PR.

PR #182, opened 2023-02-10
Update RadioBoxGroup component

luiztiago:
## Description

Updated RadioBoxGroup component following A11Y recommendations.

image

## Checklist

## Steps to Test

1. Pull down PR.
1. `npm run dev`.
1. Open Storybook.
1. Turn on the Voiceover.
1. Test the Radio Button Group component using the keyboard.

### DIFF
--- a/src/system/Form/RadioBoxGroup.js
+++ b/src/system/Form/RadioBoxGroup.js
@@ -3,76 +3,138 @@
 /**
  * External dependencies
  */
-import React from 'react';
+import React, { useCallback } from 'react';
 import PropTypes from 'prop-types';
-import * as RadioGroupPrimitive from '@radix-ui/react-radio-group';
-import { MdCheckCircle } from 'react-icons/md';
+import ScreenReaderText from '../ScreenReaderText';
 
 /**
  * Internal dependencies
  */
-import { Heading, Text } from '../';
+
+const RadioOption = ( {
+	disabled,
+	defaultValue,
+	option: { id, value, className, label, description, labelProps = {}, ...restOption },
+	name,
+	onChangeHandler,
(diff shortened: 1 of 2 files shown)


src/system/Form/RadioBoxGroup.js
```diff
@@ -3,84 +3,148 @@
...
-			sx={ {
-				display: 'flex',
-				gap: 2,
-			} }
-			ref={ forwardRef }
-			{ ...props }
+const RadioOption = ( {
+	disabled,
+	defaultValue,
+	option: { id, value, className, label, description, labelProps = {}, ...restOption },
```
djalmaaraujo, line 18: can you change id to `forLabel`? This is what we are using for all inputs and controls.

djalmaaraujo (changes requested):
Great work @luiztiago. Thank you! Just one minor thing

djalmaaraujo (approved):

Caitboard:
This fix is the way it's supposed to be done! I have one hiccup, though. There is a bug in VoiceOver, which causes issues with `<span>`s inside `<label>`s. Currently the selected radiobutton is announced as "Oneand one more item and one more item". 
image
Please see [Article about VoiceOver issue with` <span>`s in `<label>`s](https://scholar.harvard.edu/ccwilcox/blog/safari-voiceover-problem-reading-input-labels).
The description is currently inside a span with `display:block`. If I change that to `inline`, VoiceOver does announce it as it should. However this would mess up your styling. 

You could either:
- Find a way to work around this in the styling
- Use `aria-describedby` instead. So we would have a label of "One" and the input field would have an `aria-describedby` that refers to the element that contains the description.

PR #183, opened 2023-02-10
Remove letter spacing from text caps variant / table headers

luismulinari:
## Description

This PR removes the letter spacing from text caps variant / table headers.

| Before   | After |
|----------|:-------------:|
| before |  after |

## Checklist

- [N/A] This PR has good automated test coverage
- [N/A] The storybook for the component has been updated

## Steps to Test

Example:

1. Pull down PR.
2. `npm run dev`.
3. Open http://localhost:6006/?path=/story/table--default.

### DIFF
--- a/src/system/theme/textStyles.js
+++ b/src/system/theme/textStyles.js
@@ -41,6 +41,5 @@ export const textStyles = {
 		marginBottom: 2,
 		color: 'muted',
 		fontWeight: 'bold',
-		letterSpacing: '.05em',
 	},
 };


djalmaaraujo (approved):
Thanks!

donaghokeeffe:
If we use All Caps text, which we still do in places in the Dashboard, then it should still have the letter spacing applied. My input on this over in Slack was only about the table headers, because they have recently been changes away from all caps.

PR #185, opened 2023-02-15
Add role radio to RadioBoxGroup

luiztiago:
## Description

Added `role="radio"` and input label with only the option title. The description now is inside a span to keep using the same test structure as before. The voiceover is the same as before.

image

## Checklist

- [ ] This PR has good automated test coverage
- [x] The storybook for the component has been updated

## Steps to Test

1. Pull down PR.
1. `npm run dev`.
1. Open Storybook.
1. Turn on the Voiceover.
1. Test the Radio Button Group component using the keyboard.

PR #186, opened 2023-02-20
Reset Form.Autocomplete input properly when the user clears the input

andrea-sdl:
## Description

This is a Followup to #171 .

It seems like the `source` function is not called by the parent component when the user clears the input.
This PR adds an extra line of code to make sure they're in sync.

_PS: I was quite sure to have tested it in the previous PR #171 so I'm confused why it stopped working._ 
## Checklist

## Steps to Test

1. Visit http://localhost:6006/?path=/story/form-autocomplete--default&args=resetOnBlur:true (ensure you have `resetOnBlur` set to true in the controls tab
2. Select a value (search vanilla for example)
3. Delete the value with the keyboard
4. Move the focus away from the input, the selected value below the input should reset to empty.

### DIFF
--- a/src/system/NewForm/FormAutocomplete.js
+++ b/src/system/NewForm/FormAutocomplete.js
@@ -291,6 +291,7 @@ const FormAutocomplete = React.forwardRef(
 
 		useEffect( () => {
 			global.document.querySelector( `#${ forLabel }` ).addEventListener( 'blur', () => {
+				setInputQuery( global.document.querySelector( `#${ forLabel }` ).value );
 				resetInputState();
 			} );
 		}, [ forwardRef ] );


pandah3 (approved):
Works as expected, thank you 🎉 

Let's hold off on releasing a new version on top of the current version as the current version (0.32.1) includes a breaking change from #185 (see [comment](https://github.com/Automattic/vip-design-system/pull/185#issuecomment-1439266490)).

andrea-sdl:
@pandah3 thanks! I changed the PR label and will hold off releasing it ;)

pandah3:
All clear now 👍 Should be good to release!!

PR #188, opened 2023-02-25
Improve Wizard autofocus on rerenders

andrea-sdl:
## Description

Followup to #175 to improve autofocus on rerendering.
Issue was that, when rerendering, the autofocus would be triggered before it was needed (on "first" render).
To avoid this, I've moved the logic from using a ref, to using a mix of states. 
Using the `initialState` ast the starting point. 

## Checklist

## Steps to Test

see #175

### DIFF
--- a/src/system/Wizard/Wizard.js
+++ b/src/system/Wizard/Wizard.js
@@ -3,7 +3,7 @@
 /**
  * External dependencies
  */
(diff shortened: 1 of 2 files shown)


kat3samsin (approved):
Tested with https://github.com/Automattic/vip-dashboard/pull/2665 and it works! Thanks for working on this @andrea-sdl!

PR #189, opened 2023-03-06
Wizard color contrast improvements + title steps

andrea-sdl:
## Description

**This PR deprecates support for the Step title being a component. It will now only supports titles being strings.**

Changes to improve Wizard Color contrast and title steps.

### Before
CleanShot 2023-03-10 at 16 02 58@2x

### After
CleanShot 2023-03-10 at 16 03 29@2x

This PR includes
* Changing of icons to use a full circle only and change the color based on the status.
* New "Step X of Y", which is printed in the heading so that it's read/announced together with the heading itself. The text is not in full caps for[ improved readability.](https://www.a11yproject.com/posts/how-to-accessible-heading-structure/#all-caps). (screenshots were made before the change)
* Update of the colors to match the design
* Instead of getting the colors from the generic theme, we use a component-approach where the colors we need are in a single component ([see here](https://github.com/Automattic/vip-design-system/pull/189/files#diff-766d29b6d3e7340553dd59e89c2d8069a4bf232d11034f3fdacfabc50c19f319))
* A11Y: I've added a "Status: [Step completed | Current step | Step not completed]" so that screen readers users will know the status too. 
* `summary` step field support, for recapping the information of a single completed or skipped step
* "Change" button, that can be called to activate a single step
* Responsive improvements (now if you resize the screen the icon stays the same and everything is fluid)

CleanShot 2023-03-13 at 18 09 25@2x

Here it is on a small screen
CleanShot 2023-03-13 at 18 11 26@2x

## Checklist

## Steps to Test

Example:

1. Pull down PR.
1. `npm run dev`.
1. Open Storybook.
1. check the wizard component https://deploy-preview-189--vip-design-system-components.netlify.app/?path=/story/wizard--with-title-auto-focus


djalmaaraujo (changes requested):
@andrea-sdl Everything looks awesome, I think we just need to inform some aria-attributes better:

I think it's a good usage of aria-current="step": https://a11y-guidelines.orange.com/en/articles/using-aria-current-attribute/#examples-4-step-process

Also, makes me wonder if this should be a `nav`, but I would ask Caitlin from level-level to ensure after we merge, and if needed, we make a fix.

Reference: https://a11y-guidelines.orange.com/en/articles/using-aria-current-attribute/

andrea-sdl:
@djalmaaraujo thanks for the check. I played around a little bit and did 2 main changes

1. in https://github.com/Automattic/vip-design-system/pull/189/commits/2ea7ce436ae5800677a8dc12e9756fd245c29eef I hid the title and the step part from the screen readers

Why? Because if a tag is present inside the heading (even a `span`) the voiceover will announce the content by splitting into pieces, making it harder to understand.

In this way the listening experience is much improved.

2. in https://github.com/Automattic/vip-design-system/pull/189/commits/3edd35c8a0604a960a44101f5f0547c809b3be72 I added support for the `aria-current` step. There is some repetition going on, since we also include the status, but since not all screen readers might support `aria-current` I feel this is a good tradeoff for now.

djalmaaraujo (approved):
Looks awesome @andrea-sdl 
Thanks.

PR #196, opened 2023-03-16
Set label from theme-ui to use the as property

djalmaaraujo:
## Description

Use `Box` as the base component to be able to use `as` prop and other internal props.

## Steps to Test

1. Pull down PR.
2. `npm run dev`.
3. Open http://localhost:6006/?path=/story/form-label--default
4. Label is the same


brunobasto (approved):
LGTM 👍

PR #202, opened 2023-05-03
Develop

dabowman:
## Description

Just pulling in token updates from the dev branch. I also added a tool to generate responsive type styles for Figma that we might be able to automate here.


dabowman:
I'm not sure how y'all are using the token files here in the dashboard, definitely double check to make sure it's not going to affect anything there. I only make changes to the expressive token sets but syncing from dev might pull in some changes Donagh has made to the productive tokens.

dabowman:
Oh wow, I'm looking over the diff and it looks like we haven't made a PR to trunk in a long time. The current folder doesn't even have the different token sets in it, only an old version of the core set. @brunobasto is there anything to watch out for here? I don't know how Djalma was pulling tokens into the dashboard if he wasn't using the token sets we're generating for that purpose ...

PR #206, opened 2023-06-16
Input > Ability to add an adjacent component to the Input box

kat3samsin:
## Description

~Added option to add a component with `display: flex` beside the input box without affecting Label and Hint message.
This is to support adding a `copy` button beside input box.~

Created a new Input component with a Copy button > `InputWithCopyButton`

image

With notification callback
image

## Checklist

## Steps to Test

1. Navigate to the [component](https://deploy-preview-206--vip-design-system-components.netlify.app/?path=/story/form-inputwithcopybutton--default).
2. Make sure the copy button is adjacent to the input box and label 
3. Make sure text is copied

### DIFF
--- a/src/system/Form/Input.js
+++ b/src/system/Form/Input.js
@@ -27,7 +27,17 @@ const inputStyles = {
 
 const Input = React.forwardRef(
 	(
-		{ variant, label, forLabel, hasError = false, required, sx = {}, errorMessage, ...props },
+		{
+			variant,
+			label,
+			forLabel,
+			hasError = false,
+			required,
+			sx = {},
+			errorMessage,
+			flexComponent = null,
+			...props
+		},
 		ref
 	) => (
 		<React.Fragment>
@@ -36,21 +46,22 @@ const Input = React.forwardRef(
 					{ label }
 				</Label>
 			) }
-
-			<ThemeInput
-				ref={ ref }
-				id={ forLabel }
-				required={ required }
-				aria-required={ required }
-				aria-describedby={ hasError ? `describe-${ forLabel }-validation` : undefined }
-				sx={ {
-					...inputStyles,
-					...sx,
-					...( hasError ? { borderColor: 'input.border.error' } : {} ),
-				} }
-				{ ...props }
-			/>
-
+			<div sx={ flexComponent && flexComponent.display === 'flex' ? { display: 'flex' } : {} }>
+				<ThemeInput
+					ref={ ref }
+					id={ forLabel }
+					required={ required }
+					aria-required={ required }
+					aria-describedby={ hasError ? `describe-${ forLabel }-validation` : undefined }
+					sx={ {
+						...inputStyles,
+						...sx,
+						...( hasError ? { borderColor: 'input.border.error' } : {} ),
+					} }
+					{ ...props }
+				/>
+				{ flexComponent && <div sx={ { ml: 2 } }>{ flexComponent.component }</div> }
+			</div>
 			{ hasError && errorMessage && (
 				<Validation isValid={ false } describedId={ forLabel }>
 					{ errorMessage }
--- a/src/system/Form/Input.js
+++ b/src/system/Form/Input.js
@@ -35,7 +35,7 @@ const Input = React.forwardRef(
 			required,
 			sx = {},
 			errorMessage,
-			flexComponent = null,
+			flexComponent = {},
 			...props
 		},
 		ref
@@ -79,6 +79,7 @@ Input.propTypes = {
 	forLabel: PropTypes.string,
 	errorMessage: PropTypes.string,
 	sx: PropTypes.object,
+	flexComponent: PropTypes.object,
 };
 
 Input.displayName = 'Input';
--- a/src/system/Form/Input.stories.jsx
+++ b/src/system/Form/Input.stories.jsx
@@ -48,7 +48,7 @@ export const WithFlexComponent = () => (
 			flexComponent={ {
 				display: 'flex',
 				component: (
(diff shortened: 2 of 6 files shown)


src/system/Form/Input.stories.jsx
```diff
@@ -36,3 +38,23 @@ export const Default = () => (
...
+export const WithFlexComponent = () => (
+	<Form.Root>
+		<Form.Input
+			placeholder="Your input here..."
+			label="Always add a label to inputs"
+			forLabel="input-simple"
+			flexComponent={ {
+				display: 'flex',
+				component: (
+					<Button sx={ { height: '40px' } } aria-label="copy">
```
andrea-sdl: @kat3samsin not a blocker: could we add something to catch the button event, otherwise if we click it we are presented with an error page.

src/system/Form/Input.js
```diff
@@ -36,21 +46,22 @@ const Input = React.forwardRef(
...
-				aria-describedby={ hasError ? `describe-${ forLabel }-validation` : undefined }
-				sx={ {
-					...inputStyles,
-					...sx,
-					...( hasError ? { borderColor: 'input.border.error' } : {} ),
-				} }
-				{ ...props }
-			/>
-
+			<div sx={ flexComponent && flexComponent.display === 'flex' ? { display: 'flex' } : {} }>
```
andrea-sdl: @kat3samsin this works 🎉 . I was thinking if we need the extra extensibility here, since I see this being used mostly for input + copy. Should we have a dedicate component that encapsulate the input, instead of adding options to this one?

The main benefit I see in case we don't need the extra stuff and can stay with a "simple" component that is a "InputWithCopyButton", is that the declaration and the code we'll write would be probably easier to understand and we would not need to pass extra components to the Design System, since they would all belong there.

kat3samsin: This actually makes a lot of sense! I'll update to create a new component, thanks for the suggestion!

src/system/Form/Input.stories.jsx
kat3samsin: Added copy button handler in https://github.com/Automattic/vip-design-system/pull/206/commits/95f7f92ceb118eca856fa29845daa60d5ec8bbd7

src/system/Form/InputWithCopyButton.js
```diff
@@ -0,0 +1,107 @@
...
+			if ( ! ref.current ) {
+				return;
+			}
+			const originalTypeStatus = ref.current.getAttribute( 'type' ) || '';
+			if ( ref.current.getAttribute( 'value' ) === '' ) {
+				ref.current.setAttribute( 'value', '    ' );
+			}
+			ref.current.setAttribute( 'type', 'text' );
+			ref.current.select();
+			document.execCommand( 'copy' ); // eslint-disable-line no-undef
```
andrea-sdl: Looks like `execCommand` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand#browser_compatibility), should we use the [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API) instead, to ensure it's future-proof?

src/system/Form/InputWithCopyButton.js
```diff
@@ -0,0 +1,107 @@
...
+					/>
+					<div sx={ { ml: 2 } }>
+						<Button
+							sx={ { height: '40px' } }
+							aria-label={ `Copy ${ label }` }
+							onClick={ handleCopy }
+							variant="ghost"
+						>
+							<MdContentCopy sx={ { mr: 2 } } />
+							Copy
```
andrea-sdl, line 85: @kat3samsin I tested it live and I missed having some kind of feedback that the data was actually copied somewhere.
Something like "Copied !", which usually replaces the `Copy` message.

I'm not sure, though, how we should approach this from an accessibility perspective. 
Should we use`aria-atomic` so that it's announced when the content changes? 
What alternatives do you see here?

src/system/Form/InputWithCopyButton.js
andrea-sdl: We could also use React to retrieve the value as I suspect this will be mostly for read-only operations. But that's not a blocker for the PR, just an idea ;)

andrea-sdl (commented):
@kat3samsin thanks for iterating on this ❤️ I left some few more comments/ideas ;) let me know what you think!

src/system/Form/InputWithCopyButton.js
kat3samsin: @andrea-sdl thanks for bringing this up! How about adding a callback to display the notification after the copy is done?

src/system/Form/InputWithCopyButton.js
kat3samsin: Thank you for the suggestion @andrea-sdl, I will try this!

src/system/Form/InputWithCopyButton.js
andrea-sdl: @kat3samsin good idea! I like it ;)

src/system/Form/InputWithCopyButton.js
kat3samsin: Resolved in https://github.com/Automattic/vip-design-system/pull/206/commits/7fca9708bf9ee4e91841d7c8dedf378f04f921d8

src/system/Form/InputWithCopyButton.js
kat3samsin: I added the notification callback.
image

andrea-sdl (approved):
Looking good, thanks again for iterating on it ;) 🚢

PR #217, opened 2023-08-03
[PIE-3694] Apply the debounce to the source function

andrea-sdl:
## Description
Jira: https://vipjira.atlassian.net/browse/PIE-3694

When we use the debounce parameter, we were only debouncing the `handleInputChange` which is called only if we use the filtering on a predefined list through the `suggest` function.

The `handleInputChange` function is not called in case we use a `source` function to populate the autocomplete from an external set of data. 

This PR adds the debounce support to the source function so it can handle the debounce correctly when used.
## Checklist

## Steps to Test

1. Pull down PR.
1. `npm run dev`.
1. Open Storybook http://localhost:6006/?path=/story/form-autocomplete--with-slow-search-and-debounce
1. try to type `Choco`, the debounce is set to 500  and the function is set to have a 1s delay.
1. If you lower the debounce in the controls section to something like 10, and you try playing again with the text (writing `vanil` and right after that `straw`) you should see the source function being called multiple times and changing the contents because the debounce is not high enough

### DIFF
--- a/src/system/NewForm/FormAutocomplete.js
+++ b/src/system/NewForm/FormAutocomplete.js
@@ -131,6 +131,7 @@ const FormAutocomplete = React.forwardRef(
 		const [ selectedValue, setSelectedValue ] = useState( value || '' );
 		const [ inputQuery, setInputQuery ] = useState( value );
 		let debounceTimeout;
+		let sourceDebounceTimeout;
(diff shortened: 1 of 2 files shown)


djalmaaraujo (approved):
Good job!

PR #224, opened 2023-09-07
Add TypeScript

chriszarate:
## Description

Changes are well isolated in commits, reviewing commit-by-commit can be helpful.

- Add TypeScript
- Update `Box` and `Heading` components to TypeScript. This is a little tricky due to how we're wrapping Theme UI components, so we'll need to follow this pattern with other components.
- Update GitHub Action CI jobs to match other VIP projects
- Update `package.json` scripts to match other VIP projects
- Update Prettier config to match other VIP projects

## Steps to Test

Example:

1. Pull down PR.
1. `npm run dev`.
1. Open Storybook.
1. Eat cookies.


src/system/Box/Box.tsx
```diff
@@ -0,0 +1,22 @@
+/**
+ * External dependencies
+ */
+import React, { Ref } from 'react';
+import classNames, { Argument } from 'classnames';
+import { Box as ThemeBox, BoxProps } from 'theme-ui';
+
+export interface Props extends Omit< BoxProps, 'className' > {
```
abdullah-kasim: We should name it as BoxProps (so the one from theme-ui can be replaced with `BoxProps as BaseBoxProps`).

We'll have lots and lots of `Props` if we do the same for all of our other components, making import autocomplete difficult. By changing this to `BoxProps`, we'd limit ourselves to only two for each component - one provided by `design-system`, or `theme-ui`.  The autocomplete will probably prioritize `design-system` anyway because it's a direct dependency rather than a sub-dependency.

I know we have `React.ComponentProps<typeof SomeComponent>` anyway, but I'd usually rely on that only when there's no choice because type inference can fail or become tricky.

.editorconfig
```diff
@@ -0,0 +1,11 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_style = tab
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.json,*.yaml,*.yml]
```
abdullah-kasim: editorconfig now requires that we add brackets when we have comma. Not sure why - it's erroring out in PHPStorm at the very least.

```suggestion
[{*.json,*.yaml,*.yml}]
```
Or `[*.{json,yaml,yml}]`

tsconfig.json
```diff
@@ -0,0 +1,27 @@
+{
+  "compilerOptions": {
```
abdullah-kasim, line 3: Do we still do `"extends": "@tsconfig/node18/tsconfig.json"` or is that out of practice?

chriszarate: Just an oversight, added in https://github.com/Automattic/vip-design-system/pull/224/commits/5f560f6f3a6114a668b8441383e1424623e52189

Thanks!

.editorconfig
chriszarate: Done in https://github.com/Automattic/vip-design-system/pull/224/commits/1f2d5f747d4640eb4d1d46613c645eef74a01d13

src/system/Box/Box.tsx
chriszarate: Done in https://github.com/Automattic/vip-design-system/pull/224/commits/8ffa70f85da90a2e506e6534d0bb04ccff2d524e

djalmaaraujo (approved):
Running smoothly with the VIP Dashboard. I tested with a bunch of pages, and I could not find any issue

PR #230, opened 2023-09-20
Build TypeScript files and provide types

chriszarate:
## Description

The previous release simply copied over TypeScript / TSX files, which worked fine since they were built by the Dashboard's build system. But more ideally we should compile to JavaScript and include type definition files so that any project can consume it. 

This also generates type definitions for JavaScript files, mainly via React and React.PropTypes. We'll see how that goes. It could be great, or we might need to turn it off.


djalmaaraujo (approved):
Awesome, @chriszarate! Thank you!

PR #240, opened 2023-10-04
Add Toggle Slider component

saroshaga:
## Description

The name Toggle ~Text~ Slider is up for debate. We basically want a component that looks something like this:
image

So we have proposed a new component in this PR that sort of "extends" the existing `Toggle` component built on Radix UI's `Switch` 

This does need to be cleaned up aesthetically but I'm putting it up for review to get first impressions on the thinking behind it. 

## Checklist

## Steps to Test

1. Pull down PR.
1. `npm run dev`
1. `npm run storybook`
2. Open http://localhost:6006/?path=/story/toggle--toggle-text-component


aswasif007:
1. It probably needs some of that "Hawaiian tan" color (#9a6014) that we use everywhere in the dashboard. Specially to highlight the active state. cc: @donaghokeeffe 
2. Accessibility. Do we need some kind of announcer to let sighted people know what state is active? cc: @djalmaaraujo

src/system/Form/Toggle.stories.jsx
```diff
@@ -91,6 +92,20 @@ const CustomStyling = args => (
...
 
+const ToggleTextStyling = args => (
+	<form>
+		<ToggleText
+			toggleOnLabel="Table"
+			toggleOffLabel="Chart"
+			checked={ args.checked }
+			disabled={ false }
+			variant="info"
+		/>{ ' ' }
```
aswasif007: If our goal is to add something that selects `A` or `B`, we probably would be better off with a component that can be used like this:

```jsx
<ToggleText
  states={ [ "A", "B" ] }
  value="A"
  onChange={...}
/>
```

This could later be extended to add more states, if need arises. Just a thought.

saroshaga: But if the choice is always going to be binary, is there any advantage of providing an array like `states` ?

saroshaga:
> 1. It probably needs some of that "Hawaiian tan" color (#9a6014) that we use everywhere in the dashboard. Specially to highlight the active state. cc: @donaghokeeffe

I think I addressed this. I'm not sure if we need Hawaiian tan.. according to the design specs (which cannot be linked), the color is a hue of grey

> 2. Accessibility. Do we need some kind of announcer to let sighted people know what state is active? cc: @djalmaaraujo

Can we handle this by overriding the default Screen Reader Text on the VIP Dashboard ?

https://github.com/Automattic/vip-design-system/assets/14042477/20f941bc-fea7-43dd-9565-6de78815ad82

src/system/Form/Toggle.stories.jsx
t-wright: > But if the choice is always going to be binary, is there any advantage of providing an array like states ?

My two cents:

I feel like the switch component is good for setting something as enabled / disabled: `View as table: <switch here>`
But the text switch implies we're choosing from a list of options (albeit only two): `View as: "chart" | "table"` which I guess is where the states come in?

Maybe if we only want to support a binary choice we could save ourselves some effort and just use the existing switch component as is with an appropriate label to indicate one of the options is enabled or disabled: `View as table: <switch here>`?

Alternatively I had a quick look and I think styling the [RadioButtonGroup](https://vip-design-system-components.netlify.app/?path=/docs/radioboxgroup--docs) component according to the screenshot would be a light lift and makes more sense to me personally for this use-case.

Or we could add a new component to the design system based on this: https://www.radix-ui.com/primitives/docs/components/toggle-group? 

Curious what the Patisserie folks think.

chriszarate:
FYI you can use TypeScript now. See `Badge` , `Box`, `Button` and others for examples. It would be great if all new components were TypeScript.

src/system/Form/Toggle.stories.jsx
andrea-sdl: Yeah, I was wondering the same. The same component we're using for time ([the inline-select](https://vip-design-system-components.netlify.app/?path=/story/form-select--is-inline&args=label:View+as;options[0].label:Table;options[0].options:!undefined;options[0].value:table;options[1].label:Chart;options[1].options:!undefined;options[1].value:Chart)) could work here and is a much more habitual pattern than the toggle between two options. 

We would get all the accessibility done for us and the pattern is common in the dashboard (and elsewhere), so there would be no risk of confusing the user.

[The linked example will show how it'll look with the two options.](https://vip-design-system-components.netlify.app/?path=/story/form-select--is-inline&args=label:View+as;options[0].label:Table;options[0].options:!undefined;options[0].value:table;options[1].label:Chart;options[1].options:!undefined;options[1].value:Chart)
CleanShot 2023-10-05 at 15 39 19@2x

djalmaaraujo:
Thanks for pinging me for this review. I ran the code, and Storybook is complaining about the discernible text, but I have other considerations:

According to WCAG, a Switch Patten is an ON/OFF Pattern. In this case, we are not saying that Chart is a TRUE and TABLE is a false option, so that's not a valid case for a switch. Because of this, I don't think the implementation should be an evolution or another "mode" for the Toggle component, which essentially is a Switch with ON/OFF.

Reference: https://www.w3.org/WAI/ARIA/apg/patterns/switch/.

Reasons I believe it can't be a switch:

- When on, the switch element has state [aria-checked](https://w3c.github.io/aria/#aria-checked) set to true.
- When off, the switch element has state [aria-checked](https://w3c.github.io/aria/#aria-checked) set to false.

Another thing is that a switcher needs a label, and the label cannot change once you change the value. I don't see a label for this component, and if it's hidden, we are providing a different experience for sight and non-sigh users.

The screenshot below shows that this is part of a set of filters, which a Form should also encapsulate. I don't know if we did the same in other pages, but it would be the correct way. If you want to keep a form without a submit button, you must ensure that every filter change is announcing what will happen after. For example, if I change from Chart to Table, I should announce to the user that the data below is changing based on this.

image

1. If I am correct, the best solution for you is to improve our [RadioBoxGroup component](https://vip-design-system-components.netlify.app/?path=/docs/radioboxgroup--docs) (role="radiogroup"). Here's a website example that shows something similar: 

image

An example of not using visible "radio inputs": https://www.w3.org/WAI/ARIA/apg/patterns/radio/examples/radio-rating/.

2. Another solution instead of a newly designed component is a [Select component](https://vip-design-system-components.netlify.app/?path=/docs/form-select--docs).

3. A third option is to use the tabs component, move this component to the parent of the filter, and use it as a tab.

4. A fourth option (please read more about it because that's something that just came to my mind) is to use a ButtonGroup improvement. Example: https://designsystem.digital.gov/components/button-group/. The Guidance part of this page describes that button groups can be used for:

> Actions have a contextual relationship. For example, the default button group can be used when a form has a primary and alternative action.

@GaryJones Do you think this case would be valid?

--------------------------

**Ps: Please, don't merge and use this component in the VIP Dashboard without a proper A11Y review and approval. We can start with simple solutions like the one I presented if you want to "deploy and get user feedback" and iterate in a properly accessible component.**

Last thing: We added TypeScript to the project. Can you please implement using TS? 🙏

GaryJones:
> [Button group] ... Do you think this case would be valid?

No - button groups are for groups of buttons with distinct actions that may be supportive (pick bold AND italic), or opposing (backwards and forwards), and may be optional (don't add any styling).

What we want from the user is for them to pick table OR list, but they must have one selected. They can't have both, and they can't have none.

The simplest and most accurate plain HTML for "pick one and one only" is the radio input, so some variant of that should be used here.

src/system/Form/Toggle.stories.jsx
saroshaga: Thanks folks! I'm going to timebox and give the toggle group and radio button group elements a go. 

@andrea-sdl : we're already currently using the inline select but I guess I wanted to take a stab at implementing the new toggle design prescribed in the Figma

image

saroshaga: Wonder if we should just stick with this.

saroshaga:
Thanks for the valuable inputs @aswasif007 , @t-wright , @djalmaaraujo , @andrea-sdl , @GaryJones & @chriszarate ❤️  !!  I'm going to close this PR and start a new one for extending the other components as suggested by y'all. This was just a rough first pass to understand if I'm on the correct track! Conceptually, the radio button makes more sense because it is more extensible (and arguably more appropriate) than a boolean selector. However, I'm also wondering if we should just stick with the existing implementation - which is the inline select.

PR #251, opened 2023-10-11
Convert Wizard to Typescript

andrea-sdl:
## Description

This PR was intended to have only the TS conversion of the Wizard component but, as it often happens, I tried to unravel a thread of components all connected to each other.

One of the roadblocks I've encountered was the various stories, because they compose the pages with multiple components, making it harder to finish the PR.

As such, I decided to make the Wizard.stories, a little bit simpler, by using a Form.Select instead of an autocomplete.
NOTE: Before merging this PR we should check the type change are compatible with how we're using the components

## Checklist

- [X] Test that the branch won't raise type check errors when used with other projects

## Steps to Test

1. Pull down PR.
1. `npm run dev`.
1. Open Storybook.
1. Check that the components changed show no difference in the story book.


src/system/Wizard/WizardStep.tsx
```diff
@@ -53,7 +73,7 @@ const WizardStep = React.forwardRef(
 		const statusIconColor = `wizard.step.icon.${ status }`;
 		const headingColor = `wizard.step.heading.${ status }`;
 
-		const StatusIcon = complete ? BsFillCheckCircleFill : BsCircleFill;
+		const StatusIcon: IconType = complete ? BsFillCheckCircleFill : BsCircleFill;
```
chriszarate: I am a little stumped by this one, need to think about it.

andrea-sdl: Removed it. I was confused too, but I think IntelliJ was playing tricks on me. 
Running `tsc` didn't bring up errors. I removed it in https://github.com/Automattic/vip-design-system/pull/251/commits/37f0c76fb59257386d87703b9b60844601214f80

andrea-sdl: @chriszarate it's not a TS problem, but more of the 'no-unsafe-assignment' eslint rule (which was failing even with or without the type)
Adding back the IconType type doesn't really help, as for some reasons it's seeing an "any" assignment.

To fix this I've moved the ternary switch to the components part and it now passes the linting, but I'm sure will find this situation again, so if you have a better/different approach I'd love to hear it.

fix: https://github.com/Automattic/vip-design-system/pull/251/commits/f4277fa8349a8bd14fd4117b14ad3663b42ff07b

chriszarate: Nice, we were wrestling with the same issue in `Notice` and didn't think to just sidestep assignment. Thanks!

src/system/Form/Input.tsx
```diff
@@ -25,11 +24,16 @@ const inputStyles = {
...
-		{ variant, label, forLabel, hasError = false, required, sx = {}, errorMessage, ...props },
-		ref
-	) => (
+interface InputProps extends ThemeInputProps {
+	label?: string;
+	hasError?: boolean;
+	required?: boolean;
+	forLabel?: string;
+	errorMessage?: string;
+	sx?: ThemeUIStyleObject;
```
chriszarate: you can omit this, it's defined identically on `ThemeInputProps`

src/system/Form/Label.tsx
```diff
@@ -18,8 +17,16 @@ export const baseLabelStyle = {
 	lineHeight: 1.5,
 	color: baseLabelColor,
 };
+interface LabelProps {
```
chriszarate: could extend `BoxProps` here and inherit many of these props

src/system/Wizard/WizardStep.tsx
```diff
@@ -117,7 +136,7 @@ const WizardStep = React.forwardRef(
 							 * we are adding the composed title here so that it's announced correctly by the voiceover
 							 * Using tags inside the heading would make the voiceover read the heading in multiple parts
 							 **/
-							`${ stepText }: ${ title }. ${ statusText }`
+							`${ stepText }: ${ title?.toString() }. ${ statusText }`
```
chriszarate, line 139: seems like title should be a required prop

src/system/Wizard/Wizard.tsx
```diff
@@ -0,0 +1,62 @@
...
+export const Wizard = React.forwardRef< HTMLDivElement, WizardProps >(
+	(
+		{ steps, activeStep, completed = [], skipped = [], className = null, titleAutofocus = false },
+		forwardRef
+	) => {
+		const [ didMount, setDidMount ] = useState( false );
+		const [ initialStep ] = useState( activeStep );
+		// didMount helps us to track the initial render, so we can focus the title only subsequent renders
+		// to avoid stealing the focus from the page we're in.
+		useLayoutEffect( () => {
```
chriszarate, line 32: i know this is not new code, but this is confusing and duplicative. seems like a single layouteffect in WizardStep should be enough

chriszarate (approved):
looks good, some minor comments

src/system/Wizard/WizardStep.tsx
andrea-sdl: Good catch. Adding the requirement won't make that much of a difference in the code, though, since ReactNode can be null or undefined.

andrea-sdl: ReactNode also includes `| string` so we can remove that too.

src/system/Form/Input.tsx
andrea-sdl: changed in https://github.com/Automattic/vip-design-system/pull/251/commits/9618d7d8300376c3e6cd310f4cb53665ccbb5f4a

src/system/Form/Label.tsx
andrea-sdl: I'm not sure I got it right, did you mean this https://github.com/Automattic/vip-design-system/pull/251/commits/c9817a139f8e44c0ef28c490a0ec83eec5bdb8eb ? Or did you mean including all the attributes from `BoxProps`?

src/system/Wizard/WizardStep.tsx
andrea-sdl: changed in https://github.com/Automattic/vip-design-system/pull/251/commits/deb583d5edd10ec96d0a647cc46820b62274daa7

src/system/Form/Label.tsx
chriszarate: Since you are passing to `Box`, I meant a pure `extends` like

`interface LabelProps extends BoxProps {`

But if there are reasons not to do it, you can ignore this, just a tiny suggestion

andrea-sdl: done in https://github.com/Automattic/vip-design-system/pull/251/commits/0ce6d477c37878926efcd44f92a91f9a4cc654f1

chriszarate (approved):

src/system/Wizard/Wizard.tsx
andrea-sdl: @chriszarate I agree on this, but I need some more time to test this out properly (I remember it was an edge case so I want to be sure it's working as expected).

If you're ok with this I'll proceed with merging this PR so we can have all the extra components converted and get back at it in a separate one.

chriszarate: Fine with me to merge

PR #252, opened 2023-10-11
TS migration: Moves progress component to TS

iamchughmayank:
## Description

Moves progress component to TS and adds a couple of basic tests.

## Checklist

## Steps to Test

We should test that the local storybook shows the same progress bar as the prod story book.
1. pull down the PR
2. run npm run dev
3. check story book component on localhost:6006 and compare with prod story book progress component — should be same.


chriszarate (approved):

src/system/Progress/Progress.tsx
```diff
@@ -1,25 +1,33 @@
...
 import { MdCheck } from 'react-icons/md';
-import { Box, Text, Flex } from '../';
+import { Box, Text, Flex } from '..';
 
 const prefix = 'vip-progress-component';
 const uniqueID = () => Math.random().toString( 36 ).substring( 7 );
 
-const Progress = React.forwardRef(
-	( { steps, activeStep, sx, forLabel = '', className, ...props }, forwardRef ) => {
+export interface ProgressProps {
```
djalmaaraujo: I believe you can extend this from "ProgressProps" from theme-ui. It already contains sx, so you might not need to define it in the interface.

iamchughmayank: Thanks for pointing out. I have handled this in d9e18f8bc079096d4ac847977c09bbf394dc4550

PR #253, opened 2023-10-16
[PIE-3778] Improve Notice component padding and remove unused Notification component

andrea-sdl:
## Description

Jira: https://vipjira.atlassian.net/browse/PIE-3778

The Notice component icon should be aligned to the title and not to the center of the content.
This PR addresses this, plus fixes a problem with the margin, in mobile view (or in smaller alerts) where the CSS Flex would shrink the margin, making them look more packed then they should've been.

Last but not least, in this PR we're removing the Notice component.

## Before and after

| Before  | After |
| ------------- | ------------- |
| CleanShot 2023-10-16 at 11 36 53@2x  | CleanShot 2023-10-16 at 11 35 07@2x  |
| CleanShot 2023-10-16 at 11 42 20@2x  | CleanShot 2023-10-16 at 11 42 48@2x  |

## Checklist

## Steps to Test

1. Pull down PR.
1. `npm run dev`.
1. Open Storybook.
1. Check that the `Notice` component displays as expected http://localhost:6006/?path=/story/notice--default


src/system/Notice/Notice.tsx
```diff
@@ -5,43 +5,61 @@
...
+import { ThemeUIStyleObject } from 'theme-ui';
 
-const colorSystemVariant = color =>
-	( { warning: 'warning', error: 'error', alert: 'error', success: 'success', info: 'info' }[
-		color
-	] );
+interface NoticeIconProps {
+	color: string;
+	variant: keyof COLOR_VARIANTS;
+}
```
chriszarate: this can simplified with literal string union type

```suggestion
type ColorVariant = 'alert' | 'error' | 'info' | 'success' | 'warning';

interface NoticeIconProps {
	color: string;
	variant: ColorVariant;
}
```

src/system/Notice/Notice.tsx
```diff
@@ -5,43 +5,61 @@
...
 
-const NoticeIcon = ( { color, variant } ) => {
-	let Icon = MdWarning;
+export interface NoticeProps {
+	children: React.ReactNode;
+	inline?: boolean;
+	sx?: ThemeUIStyleObject;
+	title?: React.ReactNode;
+	variant?: keyof COLOR_VARIANTS;
```
chriszarate:
```suggestion
	variant?: ColorVariant;
```

src/system/Notice/Notice.tsx
```diff
@@ -5,43 +5,61 @@
...
+	info: string;
+};
+const COLOR_VARIANTS_MAPPING: COLOR_VARIANTS = {
+	warning: 'warning',
+	error: 'error',
+	alert: 'error',
+	success: 'success',
+	info: 'info',
+};
+const colorSystemVariant = ( color: keyof COLOR_VARIANTS ) => COLOR_VARIANTS_MAPPING[ color ];
```
chriszarate:
```suggestion
```

chriszarate: then remove `const systemVariant = colorSystemVariant( variant );` below and `systemVariant` => `variant`

chriszarate (approved):
nice, left one improvement

src/system/Notice/Notice.tsx
andrea-sdl: @chriszarate I think the main issue here is the `alert` that gets converted to `error`. In your example it'll mean we'll get an `alert` variant, but we want the error one. I agree though that this is not the cleanest approach. Let me think if we can implement it differently.

chriszarate: Oh right, I think we should just unwind it. Either add an `alert` variant that duplicates the styling of `error`, or update dashboard to pass `error`

donaghokeeffe:
These look good. Have we checked the dismissible one? The one with the "x" btn, just in case it causes some knock-on layout issues.

src/system/Notice/Notice.tsx
andrea-sdl: Ok, I went ahead and extended the theme so we have the `alert` color support. 
Done in https://github.com/Automattic/vip-design-system/pull/253/commits/847f53317407c15f3976524b29f32720dd03fa12

chriszarate (approved):

andrea-sdl:
> These look good. Have we checked the dismissible one? The one with the "x" btn, just in case it causes some knock-on layout issues.
@donaghokeeffe Checked them. The button is added outside of the component, thus the styling is not affected.

PR #218, opened 2023-08-08
[DONT MERGE] Rectify Dev with Trunk

dabowman:
## Description

A few sentences describing the overall goals of the pull request. Any special considerations and decisions.

Also include any references to relevant discussions and documentation.

## Checklist

- [ ] This PR has good automated test coverage
- [ ] The storybook for the component has been updated

## Steps to Test

Outline the steps to test and verify the PR here.

Example:

1. Pull down PR.
1. `npm run dev`.
1. Open Storybook.
1. Eat cookies.
1. Verify cookies are delicious.

PR #261, opened 2023-10-17
VIP Design Tokens update

djalmaaraujo:
### TODO:

- [x] Use font sizes from Figma
- [x] Use font weights from Figma
- [x] Use Border radius from Figma
- [x] Use spaces (padding, margin) from Figma
- [x] Update Input.background colors

All files inside `tokens` are pushed by Figma.

## Checklist

## Steps to Test

Please look at the Preview from Netlify; everything should work the same.


djalmaaraujo:
Reference: https://github.com/Automattic/vip-design-system/pull/218

djalmaaraujo:
@dabowman We still need to update from Figma in a future PR:

- A better way to export Shadows, it's too complex to extract
- Should we use values like `#00000005` instead of rgba?
- Export font families with the proper fallbacks?
- Toggle styles have no meaning to the existing component. Can you check the Storybook and make color updates to the Toggle styles?
- We need to rethink about line-heights being in % percentage. It can be unstable: https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#percentage. I updated the tokens, so it will update if you change to flat numbers.

kat3samsin (commented):
Thank you for working on this massive PR 🥇 

Not sure if this legit bug or is it just storybook?
image

kat3samsin (changes requested):
Thanks for the fix! I think you missed AutocompleteMulti.

kat3samsin (approved):
Awesome 🙌🏼

PR #268, opened 2023-10-19
Migrate ButtonSubmit to TypeScript

chriszarate:
## Description

Migrate ButtonSubmit to TypeScript


djalmaaraujo (approved):
@chriszarate Code looks good. Do you think we should put ButtobSubmit inside the Button docs? Or should we already create the Docs? It's fine if we don't do now.

chriszarate:
> Do you think we should put ButtobSubmit inside the Button docs? Or should we already create the Docs? 

No strong opinion. I do think it's weird that it's a separate component. Shouldn't the `loading` prop be available on `Button` too? and we could just add `type=submit`?

PR #277, opened 2023-10-26
Fix Notice behavior without title

djalmaaraujo:
### Before

image

### After

image

## Checklist

- [ ] This PR has good automated test coverage
- [x] The storybook for the component has been updated

## Steps to Test

1. Pull down PR.
2. `npm run dev`.
3. http://localhost:6006/?path=/docs/notice--docs
4. Verify Notice with one line or multiple lines have the icon positioned as left-center

PR #283, opened 2023-10-30
[PIE-3778] Readjust the NoticeIcon

andrea-sdl:
## Description

This PR is a followup to #253 (which changed the position of the NoticeIcon to be in the top) and to #277 (which fixed a bug in case the Notice component didn't have a title).

The #277 reverted the design changes, so this PR tries to address both requirements
* Having the notice icon between the title and the text, when the title is present
* Keeping the notice icon aligned to the content if we don't have the title.

There is still a use case left out, which is when the Notice width is small and/or the text of the content (without the title), splits into multiple lines.

In this case, with the changes, the icon will be centered vertical, which is not what we want (ideally we'd want the icon to be positioned as if we had a title), but I have yet to find a solution to this.

BEFORE
CleanShot 2023-10-30 at 12 13 27@2x

AFTER
CleanShot 2023-10-30 at 12 14 14@2x

## Checklist

## Steps to Test

Example:

1. Pull down PR.
1. `npm run dev`.
1. Open Storybook Notice component
1. Check that the NoticeIcon is at the top when we have the title, and centered with the content when we don't have it.

### DIFF
--- a/src/system/Notice/Notice.tsx
+++ b/src/system/Notice/Notice.tsx
@@ -30,7 +30,7 @@ export interface NoticeProps {
 type ColorVariants = 'warning' | 'error' | 'alert' | 'success' | 'info';
 
 const NoticeIcon = ( { color, variant }: NoticeIconProps ) => {
-	const sx = { color, flex: '0 0 auto', mt: 0 };
+	const sx = { color, flex: '0 0 auto' };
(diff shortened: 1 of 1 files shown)


djalmaaraujo (approved):
Sorry about the regression! Thanks for fixing.

PR #290, opened 2023-11-01
Add Nav component with some Storybook configuration

djalmaaraujo:
## Description

New Component: `Nav`.

Base Component: https://www.radix-ui.com/primitives/docs/components/navigation-menu

image

**Note:** In this first version, we support only horizontal orientation + one level of menus. In the next version, we will support two levels of menus and a vertical orientation.

If you are planning to use Next.js or any other framework, here is how it works:

```jsx
<Nav.Primary label="SSO" variant="tabs">
	{ TABS.map( tab => (
		<Nav.ItemPrimary
			key={ tab.path }
			active={ tab.path === activeTab.path }
			href={ tab.path }
			asChild
		>
			<Link
				href={ `/orgs/${ id }/sso/configurations/${ idP }/edit/${ tab.path }` }
			>
				{ tab.name }
			</Link>
		</Nav.ItemPrimary>
	) ) }
</Nav.Primary>
```

## Checklist

## Steps to Test

1. Open https://deploy-preview-290--vip-design-system-components.netlify.app/?path=/docs/navigation-nav--docs
2. Navigate using keyboard or mouse to the variants
3. The last Tabs variant looks the same as the Tabs component


chriszarate:
> If you have content inside the same page that will not affect the page Route/URL, use [Tabs](http://localhost:6006/?path=/docs/tabs--docs) component instead.

I kinda fundamentally disagree with this. _All_ navigation to discrete content within a site should affect the route / URL.

src/system/Nav/NavItem.tsx
```diff
@@ -0,0 +1,80 @@
...
+} );
+
+const NavItem = forwardRef< HTMLAnchorElement, NavItemProps >(
+	(
+		{ className, children, href, active, disabled, variant = 'primary' }: NavItemProps,
+		ref: Ref< HTMLAnchorElement >
+	) => {
+		return (
+			<NavigationMenu.Item className={ classNames( `${ VIP_NAV }-item`, className ) }>
+				<NavigationMenu.Link
```
chriszarate: we need a way to pass a framework Link component (e.g., `next/link`)

djalmaaraujo: We are working on it: https://www.radix-ui.com/primitives/docs/components/navigation-menu#with-client-side-routing. I think this should work.

chriszarate: That means we need to export `NavigationMenu.Link`, right?

djalmaaraujo: @chriszarate It already works like this:

```jsx
<Nav.Root label="SSO" variant="tabs">
	{ TABS.map( tab => (
		<Nav.Item
			key={ tab.path }
			active={ tab.path === activeTab.path }
			href={ tab.path }
			asChild
		>
			<Link
				href={ `/orgs/${ id }/sso/configurations/${ idP }/edit/${ tab.path }` }
			>
				{ tab.name }
			</Link>
		</Nav.Item>
	) ) }
</Nav.Root>

```

src/system/Nav/Nav.tsx
```diff
@@ -0,0 +1,72 @@
+/** @jsxImportSource theme-ui */
```
chriszarate, line 1: not needed
```suggestion
```

src/system/Nav/Nav.tsx
```diff
@@ -0,0 +1,72 @@
...
+		const childrenWithVariant = React.Children.map(
+			children as React.ReactElement< NavItemProps >,
+			child => {
+				if ( React.isValidElement( child ) ) {
+					return React.cloneElement( child, { variant } );
+				}
+
+				return child;
+			}
+		);
```
chriszarate: Whoa. I mean, i get it. And i'm no fan of render props. But `cloneElement` can cause problems. 

Can we just require the `variant` prop on `Nav.Item`? It's unlikely to be abused

src/system/Nav/Nav.tsx
```diff
@@ -0,0 +1,72 @@
...
+						display: 'flex',
+						listStyle: 'none',
+						justifyContent: 'flex-start',
+						m: 0,
+						px: 0,
+						flexDirection: 'row',
+						'.vip-nav-component-item:last-child': {
+							mr: 0,
+						},
+						'.vip-nav-component-item-link': variant === 'tabs' ? { px: 0, mr: 2 } : {},
```
chriszarate: Can these styles go on `Nav.Item`?

src/system/Nav/NavItem.tsx
```diff
@@ -0,0 +1,133 @@
+/** @jsxImportSource theme-ui */
```
chriszarate, line 1:
```suggestion
```

src/system/Nav/NavItem.tsx
```diff
@@ -0,0 +1,133 @@
...
+import { Theme, ThemeUIStyleObject } from 'theme-ui';
+import { NavProps } from './Nav';
+
+interface NavItemTheme extends Theme {
+	outline?: Record< string, string >;
+}
+
+export interface NavItemBaseProps extends NavigationMenu.NavigationMenuItemProps {
+	className?: string;
+	sx?: ThemeUIStyleObject | undefined;
```
chriszarate:
```suggestion
	sx?: ThemeUIStyleObject;
```

src/system/Nav/NavItem.tsx
```diff
@@ -0,0 +1,133 @@
...
+}
+
+const NavItemBase = forwardRef< HTMLLIElement, NavItemBaseProps >(
+	( { children, sx = {}, className, ...rest }: NavItemBaseProps, ref: Ref< HTMLLIElement > ) => (
+		<NavigationMenu.Item
+			className={ classNames( `${ VIP_NAV }-item`, className ) }
```
chriszarate, line 29: Dumb question, why do we accept a `className` prop on these components? Don't we want to block additional styling?

src/system/Nav/NavItem.tsx
```diff
@@ -0,0 +1,133 @@
...
+			{ ...props }
+		>
+			{ children }
+		</NavigationMenu.Link>
+	)
+);
+
+export interface NavItemProps extends NavigationMenu.NavigationMenuLinkProps {
+	className?: string;
+	disabled?: boolean | undefined;
```
chriszarate:
```suggestion
	disabled?: boolean;
```

src/system/Nav/Nav.tsx
```diff
@@ -0,0 +1,72 @@
...
+					} }
+				>
+					{ childrenWithVariant }
+				</NavigationMenu.List>
+			</NavigationMenu.Root>
+		);
+	}
+);
+
+export default Nav;
```
chriszarate: Can we avoid default exports? We've been using named exports, will probably propose a lint rule

src/system/Nav/Nav.tsx
djalmaaraujo: @chriszarate Actually, this might be related to the raw html tags issue not working, but this also makes it possible to pass `sx` to Radix components. It is not working without this.

src/system/Nav/Nav.tsx
djalmaaraujo: @chriszarate I think this is a bad experience. We would need to pass it to all `Nav.Item` this prop, which can lead to problems in the implementation, broken design, etc.

src/system/Nav/Nav.tsx
djalmaaraujo: @chriszarate I don't think I could do `:last-child` from the child perspective, can I? This has to come from the parent.

src/system/Nav/NavItem.tsx
djalmaaraujo: Same explanation as before.

src/system/Nav/NavItem.tsx
djalmaaraujo: @chriszarate All components do that. Usually, this is useful for @simonwheatley  to make his Pendo magic, etc. It also gives flexibility to the tool.

src/system/Nav/Nav.tsx
djalmaaraujo: I am open to suggestions, but here's the problem:

```jsx
<Nav.Root label="Etc">
  <Nav.Item variant="primary" href="#">PHP</Nav.Item>
  <Nav.Item variant="primary" href="#">PHP 2</Nav.Item>
</Nav.Root>
```

We need to repeat if we require it on the Item level.

src/system/Nav/Nav.tsx
chriszarate: `:last-of-type`?
https://developer.mozilla.org/en-US/docs/Web/CSS/:last-of-type

src/system/Nav/NavItem.tsx
chriszarate: Got it, but just thinking from a design governance perspective, what are the valid use cases if you remove the Pendo example (which could I think be implemented with `id` or `data-` attribute)?

src/system/Nav/Nav.tsx
chriszarate: IMO repetition is better than `cloneElement` here, but I understand the reasoning. It's marked as legacy, so I think we should avoid it. It can also cause issues with TypeScript prop types.

https://react.dev/reference/react/legacy

src/system/Nav/NavItem.tsx
djalmaaraujo: To allow id or data, you would have to allow ...props. In this case, classes is a way to go for simple alternatives. We started to include classes mainly for tracking purposes; that's it. There's no other reason, and we already have tracking set up by Simon with classes.

djalmaaraujo:
> > If you have content inside the same page that will not affect the page Route/URL, use [Tabs](http://localhost:6006/?path=/docs/tabs--docs) component instead.
> 
> I kinda fundamentally disagree with this. _All_ navigation to discrete content within a site should affect the route / URL.

@chriszarate This is related to what we discussed in a previous Slack thread, about the concept of Tabs from the WCAG perspective. It does not reflect on Route changes. And we have a Tab component for these situations. Also, this sentence speaks for something other than the Tabs component, but more for the Nav: A set of links next to each other that won't be used as a TABS (WCAG fundamentally speaking) component. Nav is a component that does route changes.

src/system/Nav/Nav.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+export const NavDisplay = forwardRef< HTMLElement, NavProps >(
+	( props: NavProps, ref: Ref< HTMLElement > ) => <Nav { ...props } variant="display" ref={ ref } />
+);
+
+export const NavLink = forwardRef< HTMLElement, NavProps >(
+	( props: NavProps, ref: Ref< HTMLElement > ) => <Nav { ...props } variant="link" ref={ ref } />
+);
+
+export const NavTab = forwardRef< HTMLElement, NavProps >(
```
chriszarate, line 60: It seems like you only need `NavPrimary` and `NavTabs` since that is the only thing `variant` is used for.

chriszarate (approved):
One suggestion but looks good. Thanks for working on this!

PR #298, opened 2023-11-06
Add new Toolbar component 

djalmaaraujo:
## Description

This component is the base for the Responsiveness Project Header Part. I will create at least two more components:

- Toolbar.Toggler (to create a hamburger menu and open the Navigation sidebar)
- Navigation sidebar (Migration from the VIP Dashboard to a generic component)

### What's included:

- Generation of the correct Valet Theme Breakpoints built-in with theme-ui;
- The new Toolbar (former Header) component and its sub-parts;
- A new variant for the `Nav` component, allowing us to re-use code 🥳 . The Toolbar can now use the `Nav` component
- Storybook examples;
- Unit tests

image

image

### The Plan

- After merging this component, I will put it in the VIP Dashboard. Once the other parts are done, we will move the other parts to the VIP Dashboard, including the Responsiveness behavior

## Checklist

## Steps to Test

1. Pull down PR. OR go to https://deploy-preview-298--vip-design-system-components.netlify.app/ and jump to item 3
2. `npm run dev`.
3. Open Storybook in Navigation / Toolbar
4. Check all components of Stories
5. Navigation works with the keyboard
6. The component is similar to the VIP Dashboard header component


src/system/Avatar/Avatar.tsx
```diff
@@ -16,22 +16,24 @@ export interface AvatarProps {
...
 		<Box
 			sx={ {
 				borderRadius: '100%',
 				height: size + 2, // +2 to compensate padding on both sides
 				width: size + 2, // +2 to compensate padding on both sides
-				border: '2px solid',
-				borderColor: isVIP ? 'primary' : 'transparent',
+				borderWidth: '2px',
+				borderStyle: 'solid',
+				borderColor: isVIP || isVIP === undefined ? 'primary' : 'transparent',
```
chriszarate, line 36: seems wrong

src/system/Toolbar/Logo.tsx
```diff
@@ -0,0 +1,40 @@
...
+import React from 'react';
+import classNames from 'classnames';
+
+/**
+ * Internal dependencies
+ */
+
+const VIP_LOGO = 'vip-logo-component';
+
+type LogoProp = {
```
chriszarate:
```suggestion
type LogoProps = {
```

src/system/Toolbar/Logo.tsx
```diff
@@ -0,0 +1,40 @@
...
+	className?: string;
+	href?: string;
```
chriszarate, line 18: looks required
```suggestion
	href: string;
```

src/system/Toolbar/Logo.tsx
```diff
@@ -0,0 +1,40 @@
...
+};
+
+export const Logo = ( { className, href, ...rest }: LogoProp ) => (
```
chriszarate: there is no rest
```suggestion
export const Logo = ( { className, href }: LogoProps ) => (
```

src/system/Toolbar/Logo.tsx
```diff
@@ -0,0 +1,40 @@
...
+import classNames from 'classnames';
+
+/**
+ * Internal dependencies
+ */
+
+const VIP_LOGO = 'vip-logo-component';
+
+type LogoProp = {
+	className?: string;
```
chriszarate, line 17: I really think we should stop accepting `className` prop unless there is a really well-defined use case that is explained in stories. It's not even clear which element the className gets applied to, it makes no sense.

src/system/Toolbar/Logo.tsx
```diff
@@ -0,0 +1,40 @@
...
+		} }
+		href={ href }
+		{ ...rest }
+	>
+		<svg
+			className={ classNames( VIP_LOGO, className ) }
+			xmlns="http://www.w3.org/2000/svg"
+			viewBox="0 0 55 24"
+			role="img"
+			aria-label="Wordpress VIP Logo"
```
chriszarate: capital p dangit
```suggestion
			aria-label="WordPress VIP Logo"
```

src/system/Toolbar/Toolbar.stories.tsx
```diff
@@ -0,0 +1,107 @@
+/**
+ * External dependencies
+ */
+import type { StoryObj } from '@storybook/react';
```
chriszarate: rebase and run linter, type imports now come last

src/system/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+import { VIP_TOOLBAR } from './index';
+import { ThemeUIStyleObject } from 'theme-ui';
+import { Flex } from '..';
+
+export type ToolbarVariant = 'primary';
+
+export interface ToolbarProps {
+	className?: string;
+	sx?: ThemeUIStyleObject;
```
chriszarate: Same with `sx` prop. What is the use case?

src/system/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,55 @@
...
+	children?: React.ReactNode;
```
chriszarate: seems required
```suggestion
	children: React.ReactNode;
```

src/system/Toolbar/ToolbarMainNav.tsx
```diff
@@ -0,0 +1,61 @@
...
+	},
+};
+
+// const responsiveStyles = {
+// 	'@media screen and (min-width: 567px)': {
+// 		display: 'flex',
+// 	},
+// };
+
+export const ToolbarMainNav = forwardRef< HTMLElement, ToolbarMainNavProps >(
```
chriszarate:
```suggestion
export const ToolbarMainNav = forwardRef< HTMLNavElement, ToolbarMainNavProps >(
```

src/system/Toolbar/ToolbarMainNav.tsx
```diff
@@ -0,0 +1,61 @@
+/** @jsxImportSource theme-ui */
+
+import React, { Ref, forwardRef } from 'react';
+
+export type ToolbarMainNavProps = {
+	label?: string;
+	children?: React.ReactNode;
```
chriszarate:
```suggestion
	children: React.ReactNode;
```

src/system/Toolbar/Logo.tsx
djalmaaraujo: @chriszarate I am still thinking about this, I think we need a `asChild` there to allow Next.js customization

src/system/Toolbar/Logo.tsx
djalmaaraujo: No rest, not life. hahaha Good catch

src/system/Toolbar/Logo.tsx
djalmaaraujo: @chriszarate Let's discuss this in P2/Slack. Otherwise, the PR will stay here forever in this discussion. Blame the classes, not the 'following-standard' PR 🙏

djalmaaraujo: https://a8c.slack.com/archives/C02T9LLBDLP/p1699363463056159

src/system/Avatar/Avatar.tsx
djalmaaraujo: It defaults to `false`.
If the user passes as `isVIP`, it will be `undefined`, so: `primary`
If the user passes as `true`, it will be `primary`
If any other value, `transparent`.

I don't believe it's wrong. Let me know how it would be better so I can adjust 🙏

src/system/Nav/Nav.stories.tsx
```diff
@@ -109,3 +114,26 @@ export const Tab: Story = {
...
+				Currently there is no use case for this variant outside the Toolbar component.
+			</p>
+
+			<Box sx={ { p: 6, backgroundColor: 'toolbar.background' } }>
+				<Nav.Toolbar sx={ { mb: 4 } } label="Nav Toolbar">
+					<NavItem.Toolbar active href="#">
+						PHP
+					</NavItem.Toolbar>
+					<NavItem.Toolbar href="https://wordpress.com">WordPress</NavItem.Toolbar>
+					<NavItem.Toolbar href="htpps://newrelic.com/">New Relic</NavItem.Toolbar>
```
chriszarate:
```suggestion
					<NavItem.Toolbar href="https://newrelic.com/">New Relic</NavItem.Toolbar>
```

src/system/Nav/Nav.stories.tsx
```diff
@@ -1,4 +1,8 @@
-import { Nav, NavItem } from '../../system';
+/**
+ * Internal dependencies
+ */
+
```
chriszarate:
```suggestion
```

src/system/Toolbar/Logo.tsx
chriszarate: Seems like you should just export the logo and let the implementor wrap in a Next.js Link?

djalmaaraujo: @chriszarate Three reasons I prefer using the way it is now:

- Wrapping with a link won't get the styles we need for the Toolbar;
- People can extend and do whatever they want with this link, wrong padding, etc;
- `as` is built-in and flexible to allow you to use Next.js links too.

chriszarate (approved):

PR #313, opened 2023-11-10
New components: Menu & Breadcrumbs

djalmaaraujo:
## Description

This component is the base for the Responsiveness Project Header Part. I will create at least two more components:

- (THIS ONE) ~Navigation sidebar (Migration from the VIP Dashboard to a generic component)~
- Breadcrumbs (Breadcrumbs with some responsive behavior)
- Toolbar.Toggler (to create a hamburger menu and open the Navigation sidebar)

This Pull Request adds another variant of the `Nav` component. It's called Menu. This variant uses the same structure as the Nav component but adds support for vertical sub-menus, similar to the existing menus of the VIP Dashboard.

### What's included

- A base structure similar to the existing sidebar of the VIP Dashboard;
- Support for sub-menus;
- AddonHeader support (header between menu groups)
- Separators

### Screenshots

image

image

### The Plan

After merging this component, I will put it in the VIP Dashboard. Once the other parts are done, we will move the other parts to the VIP Dashboard, including the Responsiveness behavior.

## Checklist

## Steps to Test — Menu

1. Open https://deploy-preview-313--vip-design-system-components.netlify.app/?path=/story/navigation-nav--menu
2. Navigate using the Keyboard
3. You can open/close
4. Items are announced as expected

## Steps to Test — Breadcrumbs

1. Open https://deploy-preview-313--vip-design-system-components.netlify.app/?path=/docs/navigation-breadcrumbs--docs
2. Navigate using the Keyboard
3. Last item is not a link
4. You can inspect the code and see a `nav`, and the last `li` with a `aria-current`
5. Use your browser to reduce the breakpoint to less than 654px, you should only see the penultimate item


kat3samsin (commented):
Amazing PR. Thank you so much @djalmaaraujo for working on this!

I noted a few things 👇🏼 

1. MenuHeader should be activated when MenuItem is active and header is collapsed. 
image


2. Safari only, Chrome is fine 
image

3. Menu are too spaced apart vertically should be 8px. Also the paddings seems to big?
image

4. There should be [animation](https://github.com/Automattic/vip-dashboard/blob/nav/replace-user-settings-menu/src/components/Navigation/SubMenu.tsx#L19) on menu expand / collapse

src/system/Nav/index.tsx
```diff
@@ -1,16 +1,26 @@
...
+	Menu,
 };
 
 export const NavItem = {
 	Primary: ItemPrimary,
 	Tab: ItemTab,
 	Toolbar: ItemToolbar,
+	Menu: ItemMenu,
+	MenuGroup: ItemGroupMenu,
+	Icon: props => <span className="vip-nav-item-icon" { ...props } />,
```
chriszarate: These "barrel files" should just contain exports, not implementation. Can you move the Icon component to the NavItem file?

src/system/Nav/NavItem.tsx
```diff
@@ -38,52 +41,91 @@ const NavItemBase = forwardRef< HTMLLIElement, NavItemBaseProps >(
...
-		</NavigationMenu.Link>
-	)
-);
-
 export interface NavItemProps extends NavigationMenu.NavigationMenuLinkProps {
 	className?: string;
 	disabled?: boolean;
 	variant?: NavVariant;
+	icon?: JSX.Element;
+	render?: React.FC< { href?: string; children?: React.ReactNode } >;
```
chriszarate: This is conflating render props and component props in a confusing way that will lead to issues. It's named like a render prop (a JavaScript function that returns JSX—not actually a component!), but it's implemented (sort of) like a component prop (a component definition that will be invoked in a JSX expression).

They are very similar in practice, and which to use probably comes down to how often the design system consumer will need to customize the props passed to the foreign link component. For example, do you expect the consumer to simply pass the Next.js `Link` component without need for further customization? If so, use a component prop, but use PascalCase to indicate that the prop expects a component definition. Also type the prop using `React.ComponentType` (don't artificially limit it to `React.FC`):

```ts
import Link from 'next/link';

<NavItem.Menu LinkComponent={ Link }>Settings</NavItem.Menu>
```

On the other hand, maybe you expect the consumer to need customization frequently (e.g., additional props passed to `Link`). If so, use a render prop, but name it more descriptively and type it as a plain function (not as a component). This will avoid the need for the consumer to create one-off components.

```ts
import Link from 'next/link';

<NavItem.Menu
   renderLink={
      ( href: string, children?: ReactNode ) => <Link children={ children } href={ href } scroll={ false } />
   }
>Settings</NavItem.Menu>
```

I will say that, in the render prop example, the "teleporting" of the `children` prop feels a little strange and unnecessary. It can be worked around, though.

The component prop pattern could be a much more straightforward implementation if you give it a default value (i.e., a generic HTML link component implementation).

chriszarate (dismissed):

src/system/Nav/NavItem.tsx
```diff
@@ -38,52 +41,91 @@ const NavItemBase = forwardRef< HTMLLIElement, NavItemBaseProps >(
...
-			{ children }
-		</NavigationMenu.Link>
-	)
-);
-
 export interface NavItemProps extends NavigationMenu.NavigationMenuLinkProps {
 	className?: string;
 	disabled?: boolean;
 	variant?: NavVariant;
+	icon?: JSX.Element;
```
chriszarate: Should `icon` also be a component prop (`Icon`)? Should the size (and maybe even icon type) be passed by this component as props? I don't really know what our long-term icon strategy is. Should this repo export an Icon component that provides all possible icon types?

djalmaaraujo:
@kat3samsin Thanks for the review!

- I updated the example with both parent and child active
- I added the animation
- Added the current margin that is similar to 8px, but it's what the design system uses

@chriszarate I might need your help. I tried to export `Breadcrumbs` as a component in the `system/index.js`, but for some reason when I do this, the `Nav` component gets broken. It gives the error "you likely forgot to export your component".

Steps to reproduce:

1 - Pull the branch
2 - Go to src/system/index.js
3 - Import and export src/system/Breadcrumbs/Breadcrumbs.tsx as Breadcrumbs
4 - Go to http://localhost:6006/?path=/story/navigation-nav--default
5 - It's broken, but the Breadcrumbs story works

The PR is still in progress, but this will be a blocker once we need to make this component available.

chriszarate:
@djalmaaraujo Works for me. How is it broken for you, specifically?

djalmaaraujo:
> @djalmaaraujo Works for me. How is it broken for you, specifically?

LOL, it works now. I think I needed to sleep, or I was exporting in the wrong way.

djalmaaraujo:
> @djalmaaraujo Works for me. How is it broken for you, specifically?

Happening again:

image

src/system/Nav/NavItem.tsx
djalmaaraujo: @chriszarate I believe the way it is now gives more flexibility. The menu is dynamic, so it would be impossible to predict all icons.

We could use the `IconComponent` property to keep the same pattern we are using now for the LinkComponent. I will update the code accordingly.

djalmaaraujo: @chriszarate I thought a little more about the subject and I went with `renderIcon` property passing `size` to the consumer. TypeScript cannot understand what's coming from IconComponent, especially when it's a customized component, like react-icon. To be safe, a `renderIcon` would work, and still be flexible.

src/system/Nav/styles.ts
```diff
@@ -75,33 +124,183 @@ export const itemVariantStyle = ( variant: NavVariant ): ThemeUIStyleObject => {
...
+				border: 'none',
+				height: 38,
+				width: '100%',
+				mb: 1,
+			};
+		}
+
+		case 'breadcrumbs': {
+			return {
+				'&::before': {
```
andrea-sdl, line 246: @djalmaaraujo minor nitpick but I think it'd be great for people jumping on this code to explain what it does in a short comment. 

This prints the `/` in the breadcrumbs, correct?
CleanShot 2023-11-21 at 16 19 52@2x

djalmaaraujo: @andrea-sdl Yes, because we should not use the separators as text, according to WCAG suggestions. I will add a comment.

andrea-sdl (commented):
@djalmaaraujo I'm trying to test it with voiceover in Safari, but seems I can't navigate with tabs or down arrow after the `Code [v]` section, it "stops" there. Which is strange. Am I doing something wrong?

I'm using this link to test https://deploy-preview-313--vip-design-system-components.netlify.app/iframe.html?id=navigation-nav--menu&viewMode=story 

I can navigate on it just fine with Chrome.

here's a video: https://github.com/Automattic/vip-design-system/assets/668251/c71f7dda-ac8a-419d-b6f4-c6caabfbc28e

andrea-sdl:
@djalmaaraujo as noted in our call, my problem was caused by this Safari setting not being set.
CleanShot 2023-11-22 at 16 30 43@2x

Everything works fine once I enable it.

andrea-sdl (approved):
Tested and it works. 🎉

PR #335, opened 2023-11-30
General Responsive Adjustments

djalmaaraujo:
## Description

We need to change some things on this side to accomplish the work in the VIP Dashboard. 

- Convert NewDialog to TS
- Created a new MenuInverse nav item
- Adjust some stories
- Moved Nav styles to their proper styles folder (refactor)
- I removed the `focus` default style in favor of `focus-visible`. All browsers now support by default so that we can achieve "only links" with focus. (https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#browser_compatibility)
- Fixed the annoying TypeScript error when using `as` with forwarded ref components. 

## Checklist

## Steps to Test — NewDialog TypeScript conversion

1. Open https://deploy-preview-335--vip-design-system-components.netlify.app/?path=/docs/dialog-newdialog--docs
2. Click on all scenarios; everything works as expected, with no regressions.

## Steps to Test — MenuInverse variant

1. Open https://deploy-preview-335--vip-design-system-components.netlify.app/?path=/story/navigation-nav--menu-inverse
2. It's a black MenuItem similar to the Toolbar item, but works vertically
3. Page should load, component rendered and that's it.

## Steps to Test — Focus fix

1. Open https://deploy-preview-335--vip-design-system-components.netlify.app/?path=/docs/navigation-link--docs
2. Focus should still happen when you keyboard focus (Press TAB), but not with the mouse


src/system/Drawer/Drawer.tsx
```diff
@@ -32,7 +35,51 @@ const Content = React.forwardRef<
...
+		DialogPrimitive.DialogContentProps {
+	children?: ReactNode;
+	trigger: ReactNode;
+	label?: string;
+	variant?: 'top' | 'right' | 'bottom' | 'left' | 'left-header' | 'right-header';
+	sx?: ThemeUIStyleObject;
+}
+
+export const Drawer = React.forwardRef<
+	React.ElementRef< typeof DialogPrimitive.Content >,
```
chriszarate: should be an HTML element, no?

chriszarate (approved):

PR #350, opened 2024-03-01
Fix wrapping of items in toolbar util menu

t-wright:
## Description

Since `1.6.5` we're extending the base nav styles in the toolbar nav which includes setting width to `100%`. This causes the util menu items (including the `VIP Features` toggle on the Dashboard) to wrap, even for large viewports. 

Before

Screenshot 2024-03-01 at 15 02 22

Screenshot 2024-03-04 at 21 21 17

After

Screenshot 2024-03-01 at 15 02 32

Screenshot 2024-03-04 at 21 22 01

## Checklist

## Steps to Test

1. When the Netlify preview link is available, visit /?path=/story/navigation-toolbar--default
1. Verify the util menu item containing whitespace is not wrapped across two lines

### DIFF
--- a/src/system/Nav/styles/variants/toolbar.ts
+++ b/src/system/Nav/styles/variants/toolbar.ts
@@ -8,6 +8,7 @@ export const toolbarRootStyles: ThemeUIStyleObject = {
 	display: [ 'none', 'none', 'flex' ],
 	height: '100%',
 	ml: 0,
+	width: 'max-content',
 };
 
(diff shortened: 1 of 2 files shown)


aswasif007 (approved):

djalmaaraujo (approved):
Good point. I have a PR that is going to be merged into the dashboard with 1.6.5. Please wait before changing something on the Dashboard.

PR #372, opened 2024-04-08
Responsive changes — General Changes + New components

### LINKED ISSUE #365: Bug: Incorrect background color used for active state of sub-nav menu item
The background color on an active submenu item does not match the color specified in Figma. It is currently using pure white. It should be using the color token **background.primary**.

Screenshot 2024-03-22 at 1 03 09 PM

djalmaaraujo:
## Description

This is a BIG Pull Request! Sorry for the big chunk of changes, but they are too integrated with some recent changes on the VIP Dashboard and other applications of the VIP Design System. I will try to be very practical on how to test this:

Fixes #365

## Checklist

### New Mobile Menu

image

1. Open https://deploy-preview-372--vip-design-system-components.netlify.app/?path=/docs/navigation-mobilemenu--docs
2. Click on the hamburger icon
3. Expect to see a RAW MobileMenu using a drawer component, with no specific paddings/margins.

### New fullWidth Button prop

image

1. Open https://deploy-preview-372--vip-design-system-components.netlify.app/?path=/docs/button--docs
2. Expect to see a full-width button

### Toggle converted to Typescript + Documentation
1. Open https://deploy-preview-372--vip-design-system-components.netlify.app/?path=/docs/toggle--docs
2. Expect this component to work as usual; there are no new features, but now it has documentation. The new primary color is dark, just in case you think it's a bug.

### Other changes
Other changes are all related to style changes.


brunobasto (approved):
Looks good to me! Thank you 🚀

src/system/Button/Button.tsx
```diff
@@ -25,11 +25,15 @@ export enum ButtonVariant {
 export interface ButtonProps extends ThemeButtonProps {
 	disabled?: boolean;
 	onClick?: ( event: ButtonClickType ) => void;
+	fullWidth?: boolean;
```
maxschmeling: Is this a common scenario we need to support with a first class API instead of just allowing `sx={ { width: '100%' } }` on the occasion that this is needed?

If so, for some reason I just don't really love the name even though I can't quite figure out why.

Thoughts on something like `stretch`?

```jsx
<Button stretch>
  Do Something
</Button>
```

src/system/Button/Button.tsx
```diff
@@ -25,11 +25,15 @@ export enum ButtonVariant {
...
 	variant?: keyof typeof ButtonVariant; // converts the enum to a string union type
 }
 
 const Button = forwardRef< HTMLButtonElement, ButtonProps >(
-	( { className, disabled, onClick, sx, ...rest }: ButtonProps, ref: Ref< HTMLButtonElement > ) => {
+	(
+		{ className, disabled, onClick, sx, fullWidth = false, ...rest }: ButtonProps,
+		ref: Ref< HTMLButtonElement >
+	) => {
```
maxschmeling: The explicit typings aren't needed here because they're already provided in the `forwardRef` signature. There also isn't really a need to provide a default for `fullWidth` since falsey is good enough for the check below.
```suggestion
	( { className, disabled, onClick, sx, fullWidth, ...rest }, ref ) => {
```

src/system/Button/Button.tsx
```diff
@@ -65,6 +69,7 @@ const Button = forwardRef< HTMLButtonElement, ButtonProps >(
 						cursor: 'not-allowed',
 						pointerEvents: 'none',
 					},
+					width: fullWidth || fullWidth === undefined ? '100%' : 'auto',
```
maxschmeling: `fullWidth` had a default set to `false` previously, so it never could have been `undefined`. Assuming you only want it to be `100%` when `fullWidth` is explicitly set to `true`, this should be good.
```suggestion
					width: fullWidth ? '100%' : 'auto',
```

src/system/Form/Toggle.tsx
```diff
@@ -0,0 +1,80 @@
+/* eslint-disable max-len */
+/** @jsxImportSource theme-ui */
+
+/**
+ * External dependencies
+ */
```
maxschmeling:
```suggestion
```

src/system/MobileMenu/MobileMenu.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { Logo } from '../Toolbar/Logo';
+import { Button, Flex, Nav, Box } from '../index';
+
+export interface MobileMenuProps {
+	children: React.ReactNode;
+	toolbarItems?: React.ReactNode;
+}
+
+export const MobileMenu = forwardRef< HTMLDivElement, MobileMenuProps >(
+	( { children, toolbarItems }: MobileMenuProps, ref: Ref< HTMLDivElement > ) => (
```
maxschmeling:
```suggestion
	( { children, toolbarItems }, ref ) => (
```

src/system/NewDialog/DialogClose.tsx
```diff
@@ -20,27 +21,39 @@ export const DialogClose = forwardRef< HTMLButtonElement, DialogCloseProps >(
...
+	color: variant === 'primary' ? 'icon.primary' : 'icon.inverse',
+	svg: {
+		'&:hover': {
+			fill: 'inherit',
+		},
+	},
+} );
+
+export const DialogCloseDefault = forwardRef< HTMLButtonElement, DialogCloseDefaultProps >(
+	( { variant = 'primary' }: DialogCloseDefaultProps, forwardedRef ) => {
```
maxschmeling:
```suggestion
	( { variant = 'primary' }, forwardedRef ) => {
```

src/system/Drawer/Drawer.tsx
```diff
@@ -4,44 +4,60 @@ import * as DialogPrimitive from '@radix-ui/react-dialog';
...
+	label?: string;
+	variant?: 'top' | 'right' | 'bottom' | 'left' | 'left-header' | 'right-header';
+	dimensions?: responsiveDimensionsProp;
+	renderClose?: () => JSX.Element | null;
+}
+
 export const Content = React.forwardRef< HTMLDivElement, DrawerContentProps >(
 	(
-		{ children, variant = 'left', label, width, height, ...rest }: DrawerContentProps,
+		{ children, variant = 'left', label, dimensions, renderClose, ...rest }: DrawerContentProps,
```
maxschmeling:
```suggestion
		{ children, variant = 'left', label, dimensions, renderClose, ...rest },
```

src/system/Form/Toggle.stories.tsx
```diff
@@ -0,0 +1,128 @@
+/** @jsxImportSource theme-ui */
+/**
+ * External dependencies
+ */
```
maxschmeling:
```suggestion
```

src/system/Form/Toggle.stories.tsx
```diff
@@ -0,0 +1,128 @@
...
+import { StoryObj } from '@storybook/react';
+
+/**
+ * Internal dependencies
+ */
```
maxschmeling:
```suggestion
```

src/system/MobileMenu/MobileMenu.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+ * External dependencies
+ */
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { axe } from 'jest-axe';
+import { ThemeUIProvider } from 'theme-ui';
+
+/**
+ * Internal dependencies
+ */
```
maxschmeling:
```suggestion
```

src/system/MobileMenu/MobileMenu.test.tsx
```diff
@@ -0,0 +1,56 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+// @ts-nocheck
+/** @jsxImportSource theme-ui */
+/**
+ * External dependencies
+ */
```
maxschmeling:
```suggestion
```

maxschmeling (approved):
Generally looks pretty great. Left a few notes / minor things to change.

src/system/Button/Button.tsx
djalmaaraujo: @maxschmeling, Good question. Ideally, we want to avoid using `sx` more and more. As we make our design system more robust, we want to avoid customization, which could break things.

As for the name, it could be just `full`, too. Stretch might infer that it could grow vertically, maybe? Not sure. I don't dislike `stretch`.

src/system/Button/Button.tsx
djalmaaraujo: It's very common in react components to pass the property as `fullWidth` without any value. We do this in other components. If I remove the false defaul value from the `fullWidth` property, and try to follow the react standard, I will have all buttons as full width:

image

That's the reason we set it to false, so we can check for undefined, which is when someone actually passes the property without a value.

Ex:

```jsx
<Button fullWidth> // correct
<Button> // correct
<Button fullWidth={true}> // correct but not the standard
<Button fullWidth={false}> // correct but you can simply omit the prop
```

I changed to a common approach to check:  Boolean( fullWidth ) === true

src/system/Button/Button.tsx
maxschmeling: I thought about `full` also. I guess either one could be taken to mean both horizontal and vertical.

I guess I mostly just question the need for it in the design system. I wonder if we could just have `grow?: boolean` that would set `flexGrow: 1` and then the button will grow to fill in the direction of it's flex container. Typically I would have a button in a div that can be a flex container.

src/system/Button/Button.tsx
maxschmeling: `<Button fullWidth>` is JSX short-hand / syntactic sugar for `<Button fullWidth={ true }>` so when you do that, the value of that prop is `true`, not `undefined`.

kat3samsin (commented):
Awesome work @djalmaaraujo! 💟 

I noticed the mobile menu names and active indicator are not aligned
SCR-20240419-htlr

This is how it should look like:
SCR-20240419-hswu

PR #388, opened 2024-05-01
Fix `jest:watch` command

brookewp:
## Description

I noticed the `jest:watch` command wasn't going into watch mode, so a quick update for that here. 

## Steps to Test

1. Pull down PR
2. `npm run jest:watch src/system/Toolbar/Toolbar.test.tsx` 
3. See that watch mode is triggered

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "format:check": "npm run cmd:format -- --check",
     "jest": "NODE_ENV=test jest --env=jsdom",
     "jest:coverage": "npm run jest && open-cli ./coverage/index.html",
-    "jest:watch": "NODE_ENV=test jest --env=jsdom --watch",
+    "jest:watch": "npm run jest -- --watch",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -15,7 +15,7 @@
     "format:check": "npm run cmd:format -- --check",
     "jest": "NODE_ENV=test jest --env=jsdom",
     "jest:coverage": "npm run jest && open-cli ./coverage/index.html",
-    "jest:watch": "npm run jest --watch",
+    "jest:watch": "NODE_ENV=test jest --env=jsdom --watch",
```
chriszarate: nice! the original is missing the `--` delimiter, so the appended arg is applied to `npm` instead of `jest`. another way to fix it would be:
```suggestion
    "jest:watch": "npm run jest -- --watch",
```

but yours is fine!

chriszarate (approved):

package.json
brookewp: Great! I had just added the envs for consistency with the `jest` command, but we don't need them there, so we can keep it simple instead. 🙂

Issue #373, opened 2024-04-17
New component: Footer


djalmaaraujo:
We have a shared Footer implementation in two of our repositories now. The picture below shows a Desktop / Mobile view of one of the applications.

<img width="859" alt="image" src="https://github.com/Automattic/vip-design-system/assets/3402/c1121884-0cea-4523-ba0f-9c888eba6855">

Currently, we also have a similar implementation in the Public login page of the Dashboard (screenshot below):

<img width="632" alt="image" src="https://github.com/Automattic/vip-design-system/assets/3402/f8bef47a-a09c-4e96-a931-059221a9175e">

There's an opportunity here to build a single component to be used in both places.

PR #390, opened 2024-05-03
LinkExternal: migrate from dashboard

brookewp:
## Summary

We're planning to migrate the footer from the dashboard (#373), which contains the  `LinkExternal` component. To make that migration easier, I figured I'd start here. 

## Details

In the first commit https://github.com/Automattic/vip-design-system/commit/0ecdce962fda7504bb8ed9e6c97394a48b26cd0a, it's 1:1 with the dashboard, just for ease of comparing the following changes in subsequent commits. But mostly, I've kept the component the same. I checked in with @djalmaaraujo to see if we wanted to use this as an opportunity to change anything about the component or if we want to keep it the same for a smoother migration. I initially thought we might want an `isExternal` prop on the existing `<Link/>` component, but I want to ensure I follow the standards set here. 🙂 

## Checklist

## Steps to Test

1. Pull down PR.
2. `npm run dev`.
3. Open Storybook.
4. Navigate to `LinkExternal` component
5. Verify component works the same as the dashboard component

### DIFF
--- a/src/system/LinkExternal/LinkExternal.tsx
+++ b/src/system/LinkExternal/LinkExternal.tsx
@@ -1,13 +1,7 @@
 /**
- * External dependencies
- */
-
-import React from 'react';
-
-import { Link } from '../Link';
-/**
  * Internal dependencies
  */
+import { Link } from '../Link';
 import ScreenReaderText from '../ScreenReaderText';
 
 // Screen reader announcements
@@ -16,12 +10,11 @@ const NEW_TAB_TEXT = ', opens in a new tab'; // reads as: link, <link text>, ext
 
 type Props = {
 	children?: React.ReactNode;
-	screenReaderText?: string | number | ReactChild;
+	screenReaderText?: string | number;
 	href: string;
 	showExternalIcon?: boolean;
 	defaultScreenReaderText?: boolean;
 	newTab?: boolean;
-	[ rest: string ]: any;
 };
 
 export const LinkExternal: React.FC< Props > = ( {
--- a/src/system/LinkExternal/LinkExternal.tsx
+++ b/src/system/LinkExternal/LinkExternal.tsx
@@ -2,6 +2,7 @@
  * External dependencies
  */
 import { translate } from 'i18n-calypso';
+
 /**
  * Internal dependencies
  */
--- a/src/system/LinkExternal/LinkExternal.tsx
+++ b/src/system/LinkExternal/LinkExternal.tsx
@@ -1,12 +1,16 @@
 /**
+ * External dependencies
+ */
+import { translate } from 'i18n-calypso';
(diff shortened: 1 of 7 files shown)


src/system/LinkExternal/LinkExternal.tsx
```diff
@@ -0,0 +1,45 @@
+/**
+ * Internal dependencies
+ */
+import { Link } from '../Link';
+import ScreenReaderText from '../ScreenReaderText';
+
+// Screen reader announcements
+const DEFAULT_EXTERNAL_LINK_TEXT = ', external link'; // reads as: link, <link text>, external link
+const NEW_TAB_TEXT = ', opens in a new tab'; // reads as: link, <link text>, external link, opens in a new tab
```
brookewp: I'm not sure why this wasn't translated in the dashboard since we have methods there from `i18n-calypso`. 

Do we not use `translate` from `i18n-calypso` in the design system to avoid having hard-coded text? Or would it make sense to implement and use here? If it's the former, we could implement the text in the dashboard using the `defaultScreenReaderText` and add translations there.

kat3samsin: Yeah it should have been translated in the VIP Dashboard. I don't think use translation in the Design System since we just pass the string through from the VIP Dashboard. We can also just start translation in the Design System if that is simpler.

kat3samsin (approved):
Looks great! Thanks for working on this. 💯

src/system/LinkExternal/LinkExternal.tsx
djalmaaraujo: I don't mind adding i18n-calypso here.

brookewp: I added in: https://github.com/Automattic/vip-design-system/pull/390/commits/6491b233146c222c7528a01ec76e6d060b32e11e

src/system/LinkExternal/index.ts
```diff
@@ -0,0 +1,4 @@
+/**
```
djalmaaraujo: We are trying to avoid index files. Can you import LinkExternal directly from the LinkExternal.tsx in src/system/index.js?

djalmaaraujo (changes requested):

src/system/LinkExternal/index.ts
brookewp: Yes, definitely!

brookewp: Made the change here: https://github.com/Automattic/vip-design-system/pull/390/commits/423eea1d22a4179a39e9be382b7bfefb5e5d62fb

djalmaaraujo (approved):
Looks great. Thanks for working on this!

PR #401, opened 2024-05-14
Responsive fixes

djalmaaraujo:
## Description

A few responsive fixes for the dashboard and general stuff:

- Making the first breaking as `0px`. This ensures we have the same behavior from 0px to the second breakpoint. Without this, the minimum breakpoint is 320, so the responsiveness is gone < 320px; 

There are just a few devices with 320px or lower, the last one being released in 2013, but it still makes sense to allow responsiveness for smaller devices.

Noting that we only need to support 320px and above since this is required for WCAG Reflow https://www.w3.org/WAI/WCAG21/Understanding/reflow.html/

- Avoid shrinking the VIP Dashboard logo and mobile trigger button;

## Checklist

## Steps to Test

These are just style changes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@automattic/vip-design-system",
-  "version": "2.4.5-dev.0",
+  "version": "2.4.5-dev.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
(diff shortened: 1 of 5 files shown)


kat3samsin (approved):
👍🏼

PR #407, opened 2024-05-30
Removed all unnecessary themes

emaurer11:
## Description

Removed unnecessary token files. This should not affect anything the Dashboard is referencing.


djalmaaraujo:
@emaurer11 Just to double check:

We use these three:

`valet-core,wpvip-product-core,wpvip-product-dark`.

We are keeping all of them, correct?

emaurer11:
@djalmaaraujo that's right! We want to keep valet-core,wpvip-product-core,wpvip-product-dark. But we should remove everything else.

djalmaaraujo (approved):

PR #409, opened 2024-06-04
Footer: Create new component 

### LINKED ISSUE #373: New component: Footer

brookewp:
## Description

Closes: https://github.com/Automattic/vip-design-system/issues/373

We need the `<Footer/>` from the dashboard login page for another app. Instead of recreating it, we can make a reusable component to use in both places. It's based on new designs: 

**Mobile**

Screenshot 2024-06-04 at 1 03 38 PM

**Desktop**

Screenshot 2024-06-04 at 1 03 22 PM

I thought it might be best to keep it as minimal as possible, so it's easy to use and keep consistent. We could always expose more props in the future, but it's not as easy taking them away 😅 

I also considered having an option for `LinkExternal` vs `Link` but then that opens it up to having the button variants (unless we don't allow it) which would require more style changes. And since LinkExternal props can be added, the arrow and screenreader text can be customized. 

## Checklist

## Steps to Test

1. Pull down PR
2. `npm run dev`
3. Open Storybook
4. Navigate to 'Footer' component
5. Ensure links work as expected, component is responsive and accessible


src/system/Footer/Footer.stories.tsx
```diff
@@ -0,0 +1,46 @@
+/**
+ * Internal dependencies
+ */
+import { StoryObj } from '@storybook/react';
+
+import { Footer } from '../Footer/Footer';
+
+export default {
+	title: 'Footer',
```
djalmaaraujo: What do you think about having `Navigation/Footer`? Since Footer is part of the navigation components, I believe it makes sense.

src/system/Footer/Footer.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+		children: 'Link1',
+		href: 'https://wpvip.com/',
+	},
+	{
+		children: 'Link2',
+		href: 'https://docs.wpvip.com/',
+	},
+];
+
+describe( '<LinkExternal />', () => {
```
djalmaaraujo: Maybe a typo? `<Footer />`

src/system/Footer/Footer.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+	it( 'should render LinkExternal props', () => {
```
djalmaaraujo: should render Footer?

src/system/Footer/Footer.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+		render( <Footer links={ combinedLinks } /> );
+
+		const link1 = screen.getByRole( 'link', { name: /link1/i } );
+		const link3 = screen.getByRole( 'link', { name: /link3/i } );
+
+		expect( link1 ).toHaveAttribute( 'target', '_self' );
+		expect( link3 ).toHaveAttribute( 'target', '_blank' );
+	} );
+
+	it( 'should hide last separator when hasTrailingSeparator is true', () => {
```
djalmaaraujo: Can you make this work with CSS using `last-child`? On the Breadcrumb components, we use the separators using CSS ::before.

Ex: https://github.com/Automattic/vip-design-system/blob/trunk/src/system/Nav/styles/variants/breadcrumbs.ts#L11

src/system/Footer/Footer.tsx
```diff
@@ -0,0 +1,104 @@
...
+	customLogo?: React.ReactNode;
+	/**
+	 * The maxiumum width of the footer.
+	 *
+	 * @default 100%
+	 **/
+	maxWidth?: string | number;
+};
+
+const a8cLogo = (
```
djalmaaraujo: Maybe a different file would fit better for this?

src/system/Footer/Footer.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+export const Footer = ( {
+	hasTrailingSeparator = false,
+	links,
+	customLogo,
+	maxWidth = '100%',
+}: FooterProps ) => {
+	return (
+		<Flex
+			sx={ {
```
djalmaaraujo, line 66: I wonder why we are not using `footer` as the HTML element here. If this is a footer, should we keep it as a landmark too?

src/system/Footer/Footer.tsx
```diff
@@ -0,0 +1,104 @@
...
+				flexDirection: [ 'column', 'column', 'row' ],
+				justifyContent: 'space-between',
+				gap: 4,
+				fontSize: 1,
+				mt: 2,
+				maxWidth,
+				textAlign: 'left',
+			} }
+		>
+			{ links && links?.length >= 0 && (
```
djalmaaraujo: Maybe checking for `links?.length` is enough here?

src/system/Footer/Footer.tsx
```diff
@@ -0,0 +1,104 @@
...
+					{ links.map( ( linkProps, index, arr ) => {
+						const hideLastSeparator = ! hasTrailingSeparator && arr.length === index + 1;
+
+						return (
+							<Box key={ `footer-link_${ index }` }>
+								<LinkExternal { ...linkProps } />
+								{ ! hideLastSeparator && (
+									<span sx={ { color: 'muted' } } aria-hidden="true">
+										/
+									</span>
```
djalmaaraujo: Read the comment about the breadcrumbs separator

src/system/Footer/Footer.tsx
```diff
@@ -0,0 +1,104 @@
...
+								) }
+							</Box>
+						);
+					} ) }
+				</Flex>
+			) }
+			<Box
+				sx={ {
+					backgroundColor: [ '#F4F3F2', '#F4F3F2', 'transparent' ],
```
djalmaaraujo: Do we have these colors on the design system? Can you check in src/system/theme/generated/valet-theme-light.json?

djalmaaraujo (changes requested):
Some minor comments. I will make other review later. Thank you!

On the preview:

image

I noticed the links are tied to the `/` separator.

src/system/Footer/Footer.stories.tsx
brookewp: I agree!

src/system/Footer/Footer.test.tsx
brookewp: Definitely! I got mixed up with the test name which I'll comment on below

src/system/Footer/Footer.test.tsx
brookewp: I'll update it to be more clear, maybe to `should accept LinkExternal props for Footer links` as I wanted to make sure that props for `LinkExternal` can be passed without error to the `Footer`

src/system/Footer/Footer.test.tsx
brookewp: Thanks for the example! I'll update to be more consistent with that 🙂

src/system/Footer/Footer.tsx
brookewp: Yes, that would make sense. I initially debated on having it as a default at all so before it was only in the story example, but then I thought it would be good for consistency. And if it reused anywhere else, would definitely be cleaner in another file. 

Do you think in the `Footer` folder or maybe in an `assets` folder?

src/system/Footer/Footer.test.tsx
brookewp: https://github.com/Automattic/vip-design-system/blob/6b6cc61a518d088b67d4b915fa21c0108006e958/src/system/Breadcrumbs/Breadcrumbs.tsx#L151-L155

@djalmaaraujo I was trying to replicate this but I'm ending up with `[Object Object]` as my class name in the dom. 😅  When looking into it, I see mentions that `sx` only works on custom components. Curious on how you made it work!

brookewp: Solved here: https://github.com/Automattic/vip-design-system/pull/409/commits/70cd4234dd163c7f6a9640e1352274d01e8e34e9

src/system/Footer/Footer.tsx
brookewp: Yes, good point! Updated here: https://github.com/Automattic/vip-design-system/pull/409/commits/e8ecf047f7ea3c4a553897addb50cd206ca4a7fd

src/system/Footer/Footer.tsx
brookewp: I went with adding an assets folder: https://github.com/Automattic/vip-design-system/pull/409/commits/3cb2ad656f115a5baf2739d50f8fc8c03f37598a

src/system/Footer/Footer.tsx
brookewp: Nice catch! Fixed: https://github.com/Automattic/vip-design-system/pull/409/commits/0cd389aab4ec04b45f31dba204e36f3e4eb92a3f

src/system/Footer/Footer.tsx
brookewp: Yes! Replaced those here: https://github.com/Automattic/vip-design-system/pull/409/commits/73ef12db404b3eb6a8b70a75aabd123c3864f73b

brookewp:
@djalmaaraujo Sorry this sat for so long! Finally I've went through your requested changes. I also added an option to show/hide link underlines since we have designs with and without: https://github.com/Automattic/vip-design-system/pull/409/commits/8235d24bebf6db5e9752a1eb322da3275f25dd37

djalmaaraujo (approved):

djalmaaraujo (approved):
@brookewp Thanks for all the changes. The only minor thing I noticed is the assets/logos.tsx file. I guess we could have a file with a better name. 

But again, this is minor.

PR #415, opened 2024-06-23
FORNO-1690: Add toggle-group component

aswasif007:
## Description

Add toggle group component.
[Click for design](https://www.figma.com/design/mLgFfiIjPIEMRQrfrJyDiN/Insights-%26-Metrics?node-id=1074-5105&t=XbenphK4b7Po1yqB-0).

Screenshot 2024-06-24 at 12 38 45 AM

## Checklist

## Steps to Test

1. Pull down PR.
1. `npm run dev`.
1. Open Storybook.
1. Eat cookies.


src/system/Form/ToggleGroup.tsx
```diff
@@ -0,0 +1,72 @@
...
+			sx={ {
+				background: 'layer.3',
+				p: 1,
+				display: 'inline-flex',
+				gap: 1,
+				borderRadius: 1,
+			} }
+		>
+			{ options?.map( option => (
+				<Button
```
t-wright: I wonder if it would be better to use actual radio buttons here and style to suit? If we use buttons, we should indicate the selected choice using `aria-checked`. There's a pattern we can follow here https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/radio_role

t-wright: Come to think of it, this component starts seeming a lot like the existing [RadioBoxGroup](https://deploy-preview-415--vip-design-system-components.netlify.app/?path=/docs/radioboxgroup--docs#default) component. Perhaps we can update that component instead, by adding a variant which is styled like our ToggleGroup? 🤔

djalmaaraujo: To @t-wright's comment: Yes. This is why we ask people to start developing components here instead of on the VIP Dashboard, so we can evaluate current options vs. creating new ones 🙏 .

The structure of the radio-group is radio under the hood. You can build a variant of the RadioBoxGroup, like @t-wright said.

src/system/Form/ToggleGroup.tsx
```diff
@@ -0,0 +1,72 @@
...
+	const handleClick = ( e: React.MouseEvent< HTMLButtonElement >, val: string ) => {
+		e.preventDefault();
+		onChange?.( val );
+	};
+
+	return (
+		<Box
+			role="group"
+			dir="ltr"
+			tabIndex={ 0 }
```
djalmaaraujo: We should never use 0 or positive values; this will affect the page tabbing order. https://www.a11yproject.com/posts/how-to-use-the-tabindex-attribute/.

Also, you are enforcing an `ltr,` which we don't usually do. Removing this line is enough.

src/system/Form/ToggleGroup.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { useState } from 'react';
+
+import { ToggleGroup } from './ToggleGroup';
+
+export default {
+	title: 'Toggle Group',
+	component: ToggleGroup,
+};
+
+export const Default = () => {
```
djalmaaraujo: Can you please add documentation for this component? Check https://vip-design-system-components.netlify.app/?path=/docs/tooltip--docs as an example on how to do it. 

Also, I believe this component should be a variant of the existing RadioBoxGroup. You can improve the documentation there.

djalmaaraujo (changes requested):
Thanks for bringing this to the vip-design-system, I really appreciate!

It looks like the implementation of this component is very similar to the RadioBoxGroup. Also, there are some concerning accessibility issues with the tabIndex implemented.

Can you check how much work would take to improve RadioBoxGroup and make this a variant of that component? Maybe a `chip` variant. I am not the best for names, but I can see a big future on extending RadioBoxGroup.

Thank you!

src/system/Form/RadioBoxGroup.js
```diff
@@ -202,6 +256,7 @@ RadioBoxGroup.propTypes = {
 	errorMessage: PropTypes.string,
 	hasError: PropTypes.bool,
 	required: PropTypes.bool,
+	variant: PropTypes.oneOf( [ 'default', 'chip' ] ),
```
aswasif007: @djalmaaraujo 
This prop is not surfacing in the documentation. Any idea why?

djalmaaraujo: @aswasif007 If I understood your question.. Maybe because RadBoxGroup is a js file?

djalmaaraujo (changes requested):
@aswasif007 Can you check Storybook A11Y issues?

image


Ps: This is a "radio" choice, so it must have a "radio" input under the hood. You should be able to navigate the keyboard keys left and right. Could you check the original RadioBoxGroup and navigate using the arrow keys?

src/system/Form/RadioBoxGroup.js
```diff
@@ -94,16 +95,65 @@ const RadioOption = ( {
...
+			sx={ {
+				background: checked ? 'layer.4' : undefined,
+				color: 'text',
+				minHeight: '32px',
+				px: 3,
+				boxShadow: checked ? 'low' : undefined,
+				fontWeight: 400,
+				fontSize: 2,
+				'&:hover': {
+					background: checked ? 'layer.4' : 'gray.3',
```
djalmaaraujo: We should not use `gray.3`. We should always reference a token, like layer.4. Can you check in Figma what the ideal token that references gray.3 should be? I believe it's `layer.1`.

src/system/Form/RadioBoxGroup.js
```diff
@@ -94,16 +95,65 @@ const RadioOption = ( {
...
+	onChangeHandler,
+} ) => {
+	const checked = `${ defaultValue }` === `${ value }`;
+
+	return (
+		<Button
+			variant="text"
+			role="radio"
+			disabled={ disabled }
+			name={ name }
```
djalmaaraujo: What is this name property in Button? It looks like you are using label already.

src/system/Form/RadioBoxGroup.js
```diff
@@ -94,16 +95,65 @@ const RadioOption = ( {
...
+	defaultValue,
+	option: { value, label },
+	name,
+	disabled,
+	onChangeHandler,
+} ) => {
+	const checked = `${ defaultValue }` === `${ value }`;
+
+	return (
+		<Button
```
djalmaaraujo: We should use the idea of an actual radio component here. To match the design, you could create the same effect using before/after pseudo-elements but still utilizing an input radio. You can check the original RadioBoxGroup for example.

src/system/Form/RadioBoxGroup.js
```diff
@@ -117,6 +167,7 @@ const RadioBoxGroup = React.forwardRef(
 			errorMessage,
 			hasError,
 			required,
+			variant = 'default',
```
djalmaaraujo: We usually use `primary` for default.

src/system/Form/RadioBoxGroup.stories.jsx
```diff
@@ -11,6 +11,36 @@ import { RadioBoxGroup } from '..';
...
+	parameters: {
+		docs: {
+			description: {
+				component: `
+A radio-box-group is a group of radio buttons that are styled as boxes. This component is used
+to allow users to select one option from a list of options.
+
+## Known issues
+
+- The <code>chip</code> variant of the current implementation of this component does not show the
```
djalmaaraujo: I don't believe it should be considered a known issue if it's not on the design.

djalmaaraujo (changes requested):

src/system/Form/RadioBoxGroup.js
aswasif007: _Also replying to:_
> You should be able to navigate the keyboard keys left and right. Could you check the original RadioBoxGroup and navigate using the arrow keys?


The issue is with keyboard navigation. If you are using radio, then you cannot focus an option with keyboard arrow-keys without making that option "active". Because moving focus to a radio option with arrow-key makes that option selected. We are going to use the chip variant as a toggle-group, to switch between chart and table modes. I think we cannot have them selected on focus, because everything else on the page will change along with it.

src/system/Form/RadioBoxGroup.js
aswasif007: To get the name as `value` in the click handler (i.e. `e.target.value`).

src/system/Form/RadioBoxGroup.js
aswasif007: Then how do the existing props surface 🤔

src/system/Form/RadioBoxGroup.js
aswasif007: Is there a better way to do it?

src/system/Form/RadioBoxGroup.js
djalmaaraujo: @ aswasif007 I understand your concern, but how you use the component does not define the component structure.

You must understand that people using the keyboard will need to use it the way you believe it's "wrong." The key to accessibility here is to provide a Radio option to customers, where they can use the tool with the keyboard. 

Suppose this component's `onChange` event will trigger a page render that depends on the code using this component. In the same way, using the keyboard can trigger the page render; a person clicking with the mouse left and right will do the same.

The UI describes a toggle situation, and we must provide an accessible switch pattern or, in this case, use a regular input radio pattern.

src/system/Form/RadioBoxGroup.js
djalmaaraujo: In the case of using the regular structure of the `Radio` component, the onChange, passes the event, and the `option` that you selected.

Here's an example:

```
const onChangeHandler = ( e: React.ChangeEvent< HTMLInputElement > ) => {
			const optionTriggered = options.find( option => {
				const optionValue = `${ option.value }`;
				const selectedOptionValue = `${ e.target.value }`;

				return optionValue === selectedOptionValue;
			} );

			if ( onChange ) {
				onChange( e, optionTriggered );
			}
		};
```

That way, you don't need value or name.

djalmaaraujo (changes requested):
I left comments in the existing threads.

src/system/Form/RadioBoxGroup.js
aswasif007: Fair enough. Updated.

src/system/Form/RadioBoxGroup.js
```diff
@@ -94,16 +94,102 @@ const RadioOption = ( {
...
+				boxShadow: checked ? 'low' : undefined,
+				'&:hover': {
+					background: checked ? 'layer.4' : 'layer.1',
+				},
+				borderRadius: 1,
+			} }
+		>
+			<input
+				ref={ ref }
+				type="radio"
```
t-wright, line 130:
```suggestion
				type="radio"
				id={ forLabel }
```
Currently we have `htmlFor={ forLabel }` on the label but it doesn't correspond to an input id. This should fix that.

t-wright (approved):
Thanks for working on this 🎉 LGTM

aswasif007:
Thanks Todd!

Djalma is on vacation so merging without his approval.

PR #417, opened 2024-06-26
Modify Card component to accept header

rebeccahum:
## Description

Modify existing Card component to accept header property:

Screenshot 2024-06-27 at 10 52 31 AM

## Checklist

## Steps to Test

Example:

1. Pull down PR.
1. `npm run dev`.
1. Open Storybook.
1. Eat cookies.

### DIFF
--- a/src/system/Card/Card.tsx
+++ b/src/system/Card/Card.tsx
@@ -1,39 +1,64 @@
+/** @jsxImportSource theme-ui */
+
 /**
  * External dependencies
  */
 import classNames, { Argument } from 'classnames';
 import { forwardRef, Ref } from 'react';
-import { BoxProps } from 'theme-ui';
+import { BoxProps, useThemeUI, get } from 'theme-ui';
 
 /**
  * Internal dependencies
  */
 import { Box } from '..';
 
+export enum CardVariant {
+	'primary',
+	'secondary',
+	'indent',
+}
+
 export interface CardProps {
-	variant?: string;
+	variant?: keyof typeof CardVariant;
 	sx?: BoxProps[ 'sx' ];
 	className?: Argument;
+	header?: string;
+	children?: React.ReactNode;
 }
 
 type CardBoxProps = CardProps & BoxProps;
 
 export const Card = forwardRef< HTMLElement, CardBoxProps >(
 	(
-		{ variant = 'primary', sx = {}, className, ...props }: CardBoxProps,
+		{ variant = 'primary', header, sx = {}, className, children, ...props }: CardProps,
 		ref: Ref< HTMLElement >
-	) => (
-		<Box
-			ref={ ref }
-			sx={ {
-				// pass variant prop to sx
-				variant: `cards.${ variant }`,
-				...sx,
-			} }
-			className={ classNames( 'vip-card-component', className ) }
-			{ ...props }
-		/>
-	)
+	) => {
+		const { theme } = useThemeUI();
+		const cardVariant = get( theme, `cards.${ variant }`, {} );
+		const basePadding = cardVariant?.p || cardVariant?.padding || 3;
+		return (
+			<Box
+				ref={ ref }
+				sx={ {
+					// pass variant prop to sx
+					variant: `cards.${ variant }`,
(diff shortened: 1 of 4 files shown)


src/system/BoxWithHeader/index.js
```diff
@@ -0,0 +1,6 @@
+/**
```
djalmaaraujo: Since this is a simple component, you don't need to import from a index file, you can import directly from the file: `export const BoxWithHeader`.

djalmaaraujo (changes requested):
First of all, thanks for creating the component here first! This is really important.

I would like to propose a different a approach from what you started, if you don't mind. The goal is to keep thins standard with the way we've be doing things.

1 - I believe the BoxWithHeading could easily be a variation of "Card". You could add a header property accepting a text or another component, and simply use the Card.

2 - The way you're customizing the variables is a little bit different of what we do usually. It might look not so "flexible", but to keep the design aligned with Figma, we usually work with "variants" of a component. Instead of passing the header bg color and header text color, you can pass the `variant="primary"` for example. The primary variant will have a "static" variation of the header you want to use.

If I may suggest what it would like in my mind, is this:

```
<Card>
  <Card.Header>This is my header</Card.Header>
  <Card.Body>This is my body</Card.Body> // this is optional
</Card>
```

You can take a look of a variant example here: src/system/Button/Button.tsx

3 - If you go with the `Card` idea, can you move this stories.tsx documentation and write some documentation? I can help too.

4 - We have a set of border colors, you might want to try `borders.2` or something similar. Figma should show the correct set of colors.

src/system/Card/Card.tsx
```diff
@@ -1,39 +1,64 @@
...
+				sx={ {
+					// pass variant prop to sx
+					variant: `cards.${ variant }`,
+					p: header ? 0 : basePadding, // Remove padding if there is a header so it sits flush
+					...sx,
+				} }
+				className={ classNames( 'vip-card-component', className ) }
+				{ ...props }
+			>
+				{ header && <header aria-label="Card header">{ header }</header> }
```
djalmaaraujo:
```suggestion
				{ header && <div>{ header }</header> }
```
While HTML5 allows the use of multiple `header` elements on a page, A11Y advises against it due to potential confusion for assistive technologies. 
Reference: https://dequeuniversity.com/rules/axe/4.2/landmark-no-duplicate-banner

The Aria role banner is similar to the header. You can use a div here; you don't need an aria-label. I think you should accept more than just a string on the header prop because this is a Heading in the end.

src/system/Card/Card.tsx
```diff
@@ -1,39 +1,64 @@
...
+	'secondary',
+	'indent',
+}
+
 export interface CardProps {
-	variant?: string;
+	variant?: keyof typeof CardVariant;
 	sx?: BoxProps[ 'sx' ];
 	className?: Argument;
+	header?: string;
```
djalmaaraujo:
```suggestion
	header?: React.ReactNode;
```

src/system/Card/Card.tsx
```diff
@@ -1,39 +1,64 @@
...
-				...sx,
-			} }
-			className={ classNames( 'vip-card-component', className ) }
-			{ ...props }
-		/>
-	)
+	) => {
+		const { theme } = useThemeUI();
+		const cardVariant = get( theme, `cards.${ variant }`, {} );
+		const basePadding = cardVariant?.p || cardVariant?.padding || 3;
```
djalmaaraujo: This seems odd to me. Can you add this default padding to the primary variant and let the variants decide the `p` style inside the card body?

It does not feel "correct" to use the card body padding inside of the `p` tag; this can lead to unexpected side effects when debugging an issue. Does it make sense?

My suggestion:

- Always use padding coming from the variant into the Box
- Let the variant style the `p`

djalmaaraujo (changes requested):
Thanks! I left some things, but it's in a much better shape now!

djalmaaraujo (approved):
Amazing! :D 
Thanks for your patience! 

:shipit:

PR #436, opened 2024-08-26
Extract the chip variant of radio-box-group to a different component

aswasif007:
## Description

Presently we have a `chip` variant of the `RadioBoxGroup` component. We are in need for a small size of this `chip` variant. But the component `RadioBoxGroup` does not provide any size prop, nor does its other variant need any.

So in this PR we are extracting the `chip` variant to a separate component `RadioGroupChip` and adding a `size` prop with `small` and `medium` values. 

output

## Steps to Test

1. Pull down PR.
1. `npm run dev`.
1. Open http://localhost:6006/?path=/docs/radiogroupchip--docs

### DIFF
--- a/src/system/Form/RadioBoxGroup.js
+++ b/src/system/Form/RadioBoxGroup.js
@@ -95,103 +95,6 @@ const RadioOption = ( {
 	);
 };
 
-const ChipOption = ( {
-	defaultValue,
-	option: { id, value, label },
-	name,
-	disabled,
-	onChangeHandler,
-} ) => {
-	const checked = `${ defaultValue }` === `${ value }`;
-	const forLabel = id || value;
-	const ref = React.useRef( null );
-	const describedById = `input-radio-box-${ forLabel }-description`;
-
-	return (
-		<div
-			id={ `o${ forLabel }` }
-			onClick={ () => {
-				ref.current?.click();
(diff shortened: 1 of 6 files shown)


djalmaaraujo (changes requested):
The implementation looks good, but you are duplicating the RadioGroup component, which already supports the Chip variation. Is there a reason we are recreating the component without removing the Chip functionality from the original?

Could adding the `size` property to the RadioBoxGroup component be a simpler solution? If you decide to go with a new component for the chip, we need to at least refactor RadioBoxGroup to remove the already-added chip feature.

PS: Since I believe you chose the "new chip component," please ensure that you reuse code from the RadioBoxGroup; otherwise, we might have different or unexpected experiences.

Milestone suggestion:

1 - Remove the current chip feature from RadioBoxGroup
2 - Release the new version of this library
3 - Open a vip-dashboard pull request with both replacements of the currently chip usage, and the new vip-design-system version

Feel free to let me know how I can help.

aswasif007:
> Is there a reason we are recreating the component without removing the Chip functionality from the original?

I will be removing the original chip variant, but was waiting on your opinion first 😄

djalmaaraujo (approved):
Thanks for working on these changes @aswasif007

PR #450, opened 2024-10-16
Update some component types to accept relevant html attributes

brookewp:
## Description

When using the following components, type errors will arise when trying to add relevant html attributes:

- Code (e.g. `id`, `name`)
- Form (e.g. `action`, `method`, `onSubmit`)
- Notice (e.g. `role`, `tabIndex`)

This PR updates the types for these three components to resolve those errors.

## Steps to Test

1. Pull down PR.
2. Edit component usages (e.g. via a story) to see that relevant attributes aren't showing type errors (e.g. `id` can be added to components noted above). 
3. Can also verify that unexpected attributes still show type error (e.g. `href` shouldn't be an option for any of the affected components noted above).


kat3samsin (approved):
Thanks for updating!

PR #451, opened 2024-10-18
Accordion: Expose props for controlling component

brookewp:
## Description

The primitive Accordion component ours is based on can be controlled or uncontrolled via `value` and `onValueChange`. However, they aren’t exposed in our component, so we get type errors as a result. 

This adds the optional types and uses existing tests to test against both versions of the component. 

## Checklist

- [ ] The storybook for the component has been updated***

***If we want to demonstrate this option, I can add a story. But if we would rather most usages be the uncontrolled option, then we should probably leave the stories as is. 

## Steps to Test

1. Pull down PR.
2. `npm run jest src/system/Accordion/Accordion.test.tsx`
3. Tests should continue to pass

### DIFF
--- a/src/system/Accordion/Accordion.test.tsx
+++ b/src/system/Accordion/Accordion.test.tsx
@@ -7,18 +7,19 @@
 import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { axe } from 'jest-axe';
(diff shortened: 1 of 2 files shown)


kat3samsin (approved):
Thanks for adding the props! Works great!

PR #460, opened 2024-11-26
Change default disabled attributes for buttons

djalmaaraujo:
## Description

Once upon a time, my old boss said: "If you don't change your mind over time, you're crazy". 
Welcome to the "putting back disabled attribute show" 😆  !

Long history short: During our new accessibility audit, we realized that adding `aria-disabled` to all `disabled` interactions might not fit all situations perfectly. For forms, we want to keep this behavior since we should always provide feedback to our users when they try to submit a form.

Some references:

- https://kittygiraudel.com/2024/03/29/on-disabled-and-aria-disabled-attributes/
- https://axesslab.com/disabled-buttons-suck/

## Checklist

## Steps to Test

1. Pull down PR.
2. `npm run dev`.
3. Open http://localhost:6006/?path=/story/button--prefer-aria-disabled
4. This rendered button has `aria-disabled` instead of disabled
5. Open http://localhost:6006/?path=/story/button--default
6. Inspect the disabled button. It has a `disabled` attribute

### DIFF
--- a/src/system/Button/Button.stories.tsx
+++ b/src/system/Button/Button.stories.tsx
@@ -119,4 +119,13 @@ const Template = args => (
 	</div>
 );
 
+const PreferAriaDisabledTemplate = args => (
+	<div>
+		<Button { ...args } preferAriaDisabled>
+			Primary
+		</Button>
+	</div>
+);
+
 export const Default = Template.bind( {} );
+export const PreferAriaDisabled = PreferAriaDisabledTemplate.bind( {} );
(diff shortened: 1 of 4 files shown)


kat3samsin (approved):
Works as expected! 

I'm curious when do you think we should use `disabled` instead of `aria-disabled`? If we are going to use `aria-disabled` more do you think we should make that as the default?

djalmaaraujo:
> Works as expected!
> 
> I'm curious when do you think we should use `disabled` instead of `aria-disabled`? If we are going to use `aria-disabled` more do you think we should make that as the default?

Thanks for asking. It used to be aria-disabled by default, but according to the audit company, if we disable a button using aria-disabled, we need to offer feedback to the user. Otherwise, the button will be "usable" but confusing to users.

I encapsulated the aria-disabled in the Form Submission since we use this component a lot, and in the case of forms, disabling would cause focus loss when loading, for example.

Issue #463, opened 2024-11-28
Feature Request: Configurable Confirmation Button ID in NewConfirmationDialog for Pendo Tracking


Crixu:
In order to improve feature usage tracking in Pendo, it is necessary to be able to identify user interactions more granularly. A critical point of interaction is the confirmation action performed by users. However, the current implementation of the NewConfirmationDialog component lacks the capability to modify the ID of the confirmation button. Tracking in Pendo works best when we can assign a tracked feature to an Element in the VIP Dashboard by it's ID. 

**Proposed Solution:**
The ability to set a custom ID for the confirmation button within the NewConfirmationDialog component upon its creation. This will allow for precise tracking of the confirmation events in Pendo.

jjyutm4532:
Hello, we tried to solve the issue.

This is what we did:

Modified the NewConfirmationDialog component to accept a new prop 'confirmButtonId' which allows setting a custom ID for the confirmation button. This change enables more granular tracking of user interactions in Pendo.

You can review changes in this commit: https://github.com/jjyutm4532/Automattic-vip-design-system-463/commit/ef10a79e861b4aee137d1ff78b3af06534e42601.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

PR #464, opened 2024-11-28
Added Possibilities to add ButtonID to Confirmation Dialog

Crixu:
## Description

This PR introduces the ability to set a custom ID or CSS selector for the confirmation button in NewConfirmationDialog. This enhancement enables better feature usage tracking in Pendo.

Related Issue: #463 

## Steps to Test

Example:

1. Pull down PR.
1. `npm run dev`.
1. Open Storybook.
1. Go to `/docs/dialog-newconfirmationdialog--docs`.
1. Verify buttonID "ConfirmIsJohnDoe" is present

### DIFF
--- a/src/system/NewConfirmationDialog/NewConfirmationDialog.js
+++ b/src/system/NewConfirmationDialog/NewConfirmationDialog.js
@@ -18,6 +18,7 @@ const NewConfirmationDialogContent = ( {
 	onConfirm,
 	onClose,
 	className = null,
+	buttonId,
 } ) => (
 	<Box className={ classNames( 'vip-confirmation-dialog-component', className ) }>
 		<Flex sx={ { justifyContent: 'flex-end', mt: 4 } }>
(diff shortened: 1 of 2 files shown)


djalmaaraujo (changes requested):
@Crixu Thanks for your contribution, but using `buttonId` is not a standard name, especially in this case that it's going to Close button, but you are passing in a parent component. This property will be confusing for users, not knowing where this ID is being applied. 

But even better would be for you to use a better CSS selector in pendo, instead of relying on IDS. Can you to the query from a upper component, like: `"#parentId button`?

Crixu:
> @Crixu Thanks for your contribution, but using `buttonId` is not a standard name, especially in this case that it's going to Close button, but you are passing in a parent component. This property will be confusing for users, not knowing where this ID is being applied.
> But even better would be for you to use a better CSS selector in pendo, instead of relying on IDS. Can you to the query from a upper component, like: `"#parentId button`?


The goal is to get an ID added to the confirm button which was achieved in this case. In the Storybook Playground I only see the ID attached to the confirm button. 
If I use the `#parentId button` CSS selector I can't differentiate between the close or the confirm button.

PR #480, opened 2025-02-21
Skeleton component

luiztiago:
## Description

- Add Skeleton Component

ezgif-1880ea1debef44

## Checklist

## Steps to Test

Example:

1. Pull down PR.
1. `npm run dev`.
1. Open [Storybook.](http://localhost:6006/?path=/docs/skeleton--docs)
1. Verify if the new component is working as expected, with 3 different variants

### DIFF
--- a/src/system/Skeleton/Skeleton.tsx
+++ b/src/system/Skeleton/Skeleton.tsx
@@ -1,0 +1,57 @@
+/**
+ * External dependencies
+ */
+
+/**
+ * Internal dependencies
+ */
+import { Box } from '../Box';
+
+export interface SkeletonProps {
+	variant?: string;
+	width?: number | string;
+	height?: number | string;
+	borderRadius?: number;
+}
+
+export const Skeleton = ( {
+	variant = 'text',
+	width = '100%',
+	height = '30px',
+	borderRadius = 1,
+	times = 1,
+	...props
+} ) => (
+	<>
+		{ Array.from( { length: times } ).map( ( i, index ) => (
+			<Box
+				key={ index }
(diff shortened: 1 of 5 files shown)


src/system/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,57 @@
...
+						'50%': {
+							opacity: 0.2,
+						},
+						'100%': {
+							opacity: 0.1,
+						},
+					},
+					mb: index === times - 1 ? 0 : 4,
+				} }
+				aria-hidden
```
rebeccahum, line 48: Are we worried about `aria-hidden` for a11y?

src/system/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,57 @@
...
+					borderRadius: variant === 'circle' ? '50%' : borderRadius,
+					width,
+					height,
+					backgroundColor: 'skeleton.background',
+					animation: 'pulse 1.5s ease-in-out infinite',
+					'@keyframes pulse': {
+						'0%': {
+							opacity: 0.1,
+						},
+						'50%': {
```
rebeccahum, line 39: Do we need to increase opacity for a11y too??

luiztiago: @rebeccahum Since it's purely decorative and not the only way to communicate progress, contrast is not strictly required. The idea is to keep with the `<Spinner />` component, and have the new component only to be a placeholder preview of content before the data gets loaded.

rebeccahum (approved):

PR #518, opened 2025-06-18
[PLTFRM-1004] Use targeted CSS properties in Label component

pandah3:
## Description

We want to change the color of the options of a checkbox form, but our current code isn't allowing us to do that (see Linear issue for screenshot).

1) **Label component:** Replace `all: 'unset'` to specific CSS properties instead since `all: 'unset'` resets any custom CSS we pass down, making it just default to the default color.

2) **Checkbox component:** Remove the hardcoded color for labels in the Checkbox component so it doesn't override the custom CSS colors we pass down to the `<Label>` component

3) **Label component Storybook:** Update the Storybook for the `<Label>` component so we can test our changes on the UI.

| BEFORE |
| ------------- | 
|Screenshot 2025-06-17 at 6 03 02 PM|

| AFTER |
| ------------- | 
|Screenshot 2025-06-17 at 6 05 38 PM|

## Checklist

### Test with local Dashboard:
Follow instructions in this PR: https://github.com/Automattic/vip-dashboard/pull/4519

### Test only the design system bit:
1. Pull down PR.
1. `npm run dev`.
1. Go to: http://localhost:6006/?path=/story/form-label--default
1. Controls tab > Add this to the `sx` field:
```
{
  "color": "texts.secondary"
}
```
1. Inspect the label and make sure:
- Color is `--theme-ui-colors-texts-secondary` and not `--theme-ui-colors-input-text-default`
- Margin/padding/border/outline is all `0`

### DIFF
--- a/src/system/Form/Checkbox/styles.ts
+++ b/src/system/Form/Checkbox/styles.ts
@@ -36,7 +36,6 @@ export const checkboxStyle = ( variant: string ): ThemeUIStyleObject => {
 		},
 		'& ~ label': {
 			fontWeight: 'regular',
-			color: inputBaseText,
 			m: 0,
 			ml: 2,
 		},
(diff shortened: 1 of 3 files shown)


kat3samsin (approved):
🚢

PR #635, opened 2026-05-26
[VIP-2897] Add inline-chips variant for AutocompleteMulti

aswasif007:
## Summary
https://linear.app/a8c/issue/VIP-2897/add-variant-to-autocompletemulti

Adds a new `variant="inline-chips"` for the `AutocompleteMulti` component. Selected items render as removable chips inside the input container instead of below it, matching the design spec from Emily's P2 post.

### Key behaviors
- Chips display inline inside the bordered input area with the text field
- Dropdown shows all options with ✓ checkmarks for selected items
- Clicking a selected option toggles it off (deselects)
- Dropdown stays open after each selection for rapid multi-select
- Placeholder hidden when items are selected
- Full keyboard navigation and screen reader support

### Usage

```jsx
<Form.AutocompleteMulti
  variant="inline-chips"
  label="Post Categories"
  options={categories}
  showAllValues
  placeholder="Search categories..."
/>
```

## Screenshot

Screenshot 2026-05-26 at 7 47 29 PM

## Changes

- **New**: `FormAutocompleteMultiselectInlineChip.tsx` — chip component (pill + close button)
- **Modified**: `FormAutocompleteMultiselect.jsx` — `variant` prop, inline-chips layout, toggle selection, keep-menu-open behavior, removal announcements, focus management
- **Stories**: `InlineChips` story with pre-populated categories
- **Tests**: 6 tests total (2 existing + 4 new for inline-chips rendering, chip removal, and a11y announcements)

## Accessibility

- Chips have `aria-label="Remove {option}"` on close buttons
- Screen reader announces both additions and removals via `aria-live="assertive"`
- Focus moves to input after chip removal
- `jest-axe` accessibility test passes

## Checklist

## Steps to Test

1. Pull down PR
2. `npm run dev`
3. Open Storybook → Form/AutocompleteMulti → InlineChips story
4. Verify chips render inside the input container
5. Click dropdown arrow → verify all options visible with checkmarks on selected
6. Select an option → verify chip added, dropdown stays open
7. Click selected option → verify chip removed (toggle), dropdown stays open
8. Click × on chip → verify chip removed
9. Type to filter → verify dropdown filters options
10. Verify existing stories (Default, WithBadges, WithStaticData) still work
11. `npm run jest` — all 6 tests pass

### DIFF
--- a/src/system/NewForm/FormAutocompleteMultiselect.jsx
+++ b/src/system/NewForm/FormAutocompleteMultiselect.jsx
@@ -106,73 +106,32 @@ const searchIconStyles = {
 };
 
 const inlineChipsContainerStyles = {
-	...baseBorderTextColors,
-	width: '100%',
+	...defaultStyles,
 	display: 'flex',
 	flexWrap: 'wrap',
 	alignItems: 'center',
 	p: 1,
-	pr: 7,
+	pr: 0,
 	position: 'relative',
-	minHeight: '36px',
+	'& .autocomplete__input': {
+		...defaultStyles[ '& .autocomplete__input' ],
+		lineHeight: '24px',
+		minHeight: '24px',
+	},
 	'&:focus-within': theme => theme.outline,
 	'& .autocomplete__wrapper': {
 		position: 'static',
 		width: '100%',
-	},
-	'& .autocomplete__input': {
-		width: '100%',
-		paddingLeft: 2,
-		py: 0,
-		borderWidth: 0,
-		color: 'text',
-		bg: 'transparent',
-		minHeight: '28px',
-		lineHeight: '28px',
-		'&:focus-visible': { outlineWidth: 0, boxShadow: 'none' },
-		'&:focus-within': { outlineWidth: 0, boxShadow: 'none' },
-		'&.autocomplete__input--focused': { outlineWidth: 0, boxShadow: 'none' },
-		'&.autocomplete__input--show-all-values': { paddingRight: 7 },
-		'&::placeholder': {
-			color: 'input.text.placeholder',
-			opacity: 1,
+		lineHeight: '24px',
+		minHeight: '24px',
+		'& .autocomplete__dropdown-arrow-down': {
+			top: 'unset',
+			bottom: '6px',
 		},
-	},
-	'& .autocomplete__menu': {
-		...baseBorderTextColors,
-		position: 'absolute',
-		left: '-1px',
-		right: '-1px',
-		top: '100%',
-		width: 'auto',
-		zIndex: 100,
-		boxShadow: 'rgba(0, 0, 0, 0.257) 0px 2px 6px',
-	},
-	'& .autocomplete__hint, & .autocomplete__input, & .autocomplete__option': {
-		fontSize: 'inherit',
-	},
-	'& .autocomplete__option': {
-		borderColor: baseControlBorderStyle.borderColor,
-	},
-	'& .autocomplete__option--odd': {
-		bg: 'layer.1',
-	},
-	'& .autocomplete__option:hover, & .autocomplete__option--focused': {
-		bg: 'input.background.primary',
-		borderColor: 'input.background.primary',
-	},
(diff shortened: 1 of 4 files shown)


src/system/NewForm/FormAutocompleteMultiselect.jsx
```diff
@@ -283,11 +354,42 @@ const FormAutocompleteMultiselect = React.forwardRef(
...
+			},
+			[ selectedOptions, unselectValue ]
+		);
+
+		const inlineChipsTemplates = useMemo(
+			() => ( {
+				suggestion: suggestion => {
+					const isSelected = selectedOptions.includes( suggestion );
+					const check = isSelected ? '&#10003;' : '';
+					return `<span style="display:flex;align-items:center;gap:8px"><span style="width:16px;flex-shrink:0">${ check }</span>${ suggestion }</span>`;
```
andrea-sdl: @aswasif007 This can be a security issue because `accessible-autocomplete` uses `dangerouslySetInnerHTML` [for the suggestions part](https://github.com/alphagov/accessible-autocomplete/blob/v2.0.4/src/autocomplete.js#L537), so theoretically they could be used to run stuff from the sites into the VIP Dashboard.

andrea-sdl (changes requested):
@aswasif007 thank you for this!
Two things I've found while testing

1. I can't trigger the menu if I click on the right part.
https://github.com/user-attachments/assets/c93201da-c423-4339-b82a-7543b18335e0

2. The implementation differs from the design  (dunno if that's expected)
Image

3 (noted in comments) there's a possible security issue we need to figure out :)

aswasif007:
> The implementation differs from the design (dunno if that's expected)

The change in the dropdown UI requires broader change that effects all other variants. I am not sure if we should do that here.

src/system/NewForm/FormAutocompleteMultiselect.jsx
aswasif007: Fixed here: https://github.com/Automattic/vip-design-system/pull/635/changes/f49f89364fa22bdc0fb980fcb9717443ff2b8712

aswasif007: Fixed in 741c8213. Added a container-level `onClick` handler that forwards clicks to the input element. When the arrow's `pointerEvents: 'none'` lets the click pass through, it now reaches the container handler which programmatically focuses and clicks the input, triggering the dropdown to open.

Also added `escapeHtml()` for the suggestion template in the previous commit (f49f8936) to address the XSS concern.

aswasif007:
> I can't trigger the menu if I click on the right part.

Fixed here: https://github.com/Automattic/vip-design-system/pull/635/commits/f8641cc0c6ce8e6d56aec62fd5019fa13e8cf6e9

src/system/NewForm/FormAutocompleteMultiselect.jsx
```diff
@@ -283,11 +321,44 @@ const FormAutocompleteMultiselect = React.forwardRef(
...
-			[ autoFilter, isDirty, onInputChange, options, selectedOptions ]
+			[ selectedOptions, unselectValue ]
+		);
+
+		const inlineChipsTemplates = useMemo(
+			() => ( {
+				suggestion: suggestion => {
+					const isSelected = selectedOptions.includes( suggestion );
+					const check = isSelected ? '&#10003;' : '';
+					return `<span style="display:flex;align-items:center;gap:8px"><span style="width:16px;flex-shrink:0">${ check }</span>${ escapeHtml(
```
andrea-sdl, line 356: one small nitpick here, dunno if that's feasible though: I wonder if we could extract the style CSS so we can get some type hints if we'll need to update these.Not a big deal though.

andrea-sdl (approved):

src/system/NewForm/FormAutocompleteMultiselect.jsx
aswasif007: will address in a follow up pr

PR #641, opened 2026-06-04
AutocompleteMulti design polish

aswasif007:
## Summary

We are making two changes here:

1. Remove the rule next to the chevron
2. Align the chevron with the top badge row on wrap

## Changes

Before | After
--- | ---
Screenshot 2026-06-04 at 1 15 41 PM | Screenshot 2026-06-04 at 1 15 45 PM

## Testing

1. Run the PR with `npm run dev`
2. Go to http://localhost:6006/?path=/story/form-autocompletemulti--inline-chips and test out

### DIFF
--- a/src/system/NewForm/FormAutocompleteMultiselect.jsx
+++ b/src/system/NewForm/FormAutocompleteMultiselect.jsx
@@ -111,7 +111,7 @@ const inlineChipsContainerStyles = {
 	flexWrap: 'wrap',
 	alignItems: 'center',
 	p: 1,
-	pr: 0,
+	pr: 5,
 	position: 'relative',
 	'& .autocomplete__input': {
 		...defaultStyles[ '& .autocomplete__input' ],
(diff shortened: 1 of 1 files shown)


src/system/NewForm/FormAutocompleteMultiselect.jsx
```diff
@@ -446,7 +448,7 @@ const FormAutocompleteMultiselect = React.forwardRef(
 								unselectValue={ unselectValue }
 							/>
 						) ) }
-						<div sx={ { flex: '1 1 120px', minWidth: '120px' } }>
+						<div sx={ { flex: '1 1 120px', minWidth: '120px', mr: -5 } }>
```
andrea-sdl, line 451: nitpick: I'm not a super fan of using negative margins, but that's me :)

andrea-sdl (approved):
Looks good. left a small comment only because I'm not a fan of negative margins, but this is good to go.

src/system/NewForm/FormAutocompleteMultiselect.jsx
aswasif007: me neither, but I had to put it because the spacing of this variant is conflicting with spacing of other variants. so it is either put a negative margin, or do a top-down refactor. 🤷

PR #689, opened 2026-08-17
Add ServiceHeader and a dismissible Notice

aswasif007:
## Description

We are making two additions:

- `ServiceHeader` component:
This is the [standard header component](https://pixelsandpointsp2.wordpress.com/2026/08/11/vip-dashboard-service-header-design-pattern-i1/) that is used in the VIP APM and Defensive Mode pages. We are porting it out to VIP Design Systems. 

Screenshot 2026-08-17 at 6 09 45 PM
Screenshot 2026-08-17 at 6 09 57 PM

- Dismiss button in `Notice` component:
We are making the `Notice` component dismissible by adding a `dismissible` prop and cross icon.

Screenshot 2026-08-17 at 6 10 09 PM

## How to test

1. Pull the PR and run `pnpm dev`
2. Go to http://localhost:6006/?path=/story/notice--dismissible&args=dismissible:!true and observe the dismissible button on the Notice component.
3. Go to http://localhost:6006/?path=/docs/serviceheader--docs and observe the new ServiceHeader component.

### DIFF
--- a/src/system/Notice/Notice.tsx
+++ b/src/system/Notice/Notice.tsx
@@ -5,9 +5,10 @@
  */
 import * as Collapsible from '@radix-ui/react-collapsible';
 import classNames from 'classnames';
+import { useTranslate } from 'i18n-calypso';
 import React, { useState, useId } from 'react';
 import { BiChevronDown, BiChevronUp } from 'react-icons/bi';
-import { MdError, MdWarning, MdCheckCircle, MdInfo } from 'react-icons/md';
+import { MdError, MdWarning, MdCheckCircle, MdInfo, MdClose } from 'react-icons/md';
 import { ThemeUIStyleObject } from 'theme-ui';
 
 /**
@@ -45,6 +46,14 @@ export type NoticeProps = Omit< React.HTMLAttributes< HTMLDivElement >, 'title' 
 	headingVariant?: React.ElementType;
 	/** Additional CSS class name(s) appended to the root element. */
 	className?: string;
+	/**
+	 * When true, renders a close button that hides the notice. Ignored on a
+	 * collapsible notice, which shows only its toggle: two icons in one header
+	 * leave the chevron ambiguous. @default false
+	 */
+	dismissible?: boolean;
+	/** Called when the notice is dismissed through the close button. */
+	onDismiss?: () => void;
 	/** Ref forwarded to the root element. */
 	ref?: React.Ref< HTMLDivElement >;
 } & CollapsibleProps;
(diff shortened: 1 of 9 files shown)


src/system/Notice/Notice.tsx
```diff
@@ -45,6 +46,10 @@ export type NoticeProps = Omit< React.HTMLAttributes< HTMLDivElement >, 'title'
 	headingVariant?: React.ElementType;
 	/** Additional CSS class name(s) appended to the root element. */
 	className?: string;
+	/** When true, renders a close button that hides the notice. @default false */
+	dismissible?: boolean;
```
andrea-sdl, line 54: I do wonder if the notice could either be dismissible or collapsible because once the two icons are there it makes it harder to understand if the collapsible icon is cosmetic or not, as opposed to when it's the only one, in which case it's clear that it's an interactive element. 

Image

andrea-sdl (approved):
Left a small comment, but I think the overall work is solid. Personally I would favor not allowing dismissable+collapsible, this way the icons are clearer to the user when present.

donaghokeeffe:
Sorry, came late to this. 

1. Agree with Andrea's comment about either collapsable or dismissable, but not both. 
2. Min target size for the close button is 32px x 32px (for touch screen devices).
3. The hover state of the close button needs some tweaks. To avoid giving it a new hover color per notice type I'd recommend using a blend mode to make the hover color looks like a darker version of he current notice's bg color. The following css should work if placed on the close button's hover state: 

```
background-color: rgb(128 128 128 / 30%);
mix-blend-mode: multiply;
```
  
 4. Finally, don't make any other states of the service header for the moment, because that was the first draft design and I'm planning on a second. What's shown above in this PR is sufficient for the APM use case for now anyway. 

-------- 
 
I was doing some QA work on the VIP APM UI today and Claude highlighted a few issues with the service header. I'm pasting them here for your info. 
 
 **Claude says:**

> Four things I'd flag on src/system/ServiceHeader/ServiceHeader.tsx, all carried over from the dashboard version rather than
>   introduced here.
> 
>   1. Status badge has no ID hook
> 
>   The <Badge> only takes variant and sx, so there's no way to attach a Pendo identifier. Our Pendo checklist in AGENTS.md calls this out specifically:
> 
>   ▎ Status indicators (badges, labels) have IDs for visibility tracking
> 
>   Every interactive control on the VIP APM page follows vip-{feature}-{action}-{element} (vip-apm-enable-button, vip-apm-open-button), but the status
>   badge doesn't, so we can't measure which state users actually land in — which for a service in open beta is probably the most useful signal we're
>   missing.
> 
>   Since this is now a shared component it can't hardcode an APM-specific ID. Could we add a badgeId prop (or spread ...rest onto the badge) so consumers
>   pass their own? Worth doing now — retrofitting later means touching every consumer.
> 
>   2. aria-live is scoped wider than the thing that changes
> 
>   <Box aria-live="polite">
>     <Heading>{ title }</Heading>
>     <Badge>…{ badge.label }</Badge>
>     <Text>{ description }</Text>
>   </Box>
> 
>   Only the badge ever changes, but the live region wraps the title and description too, so every status change re-announces the entire block — title,
>   status, and the full service description. Moving aria-live="polite" onto the Badge would announce just the part that changed.
> 
>   3. Decorative icon missing aria-hidden
> 
>   <BiInfoCircle size={ 14 } sx={{ fill: 'texts.helper' }} /> in the message bar is decorative, but unlike its siblings in the same file (Spinner and
>   FaCircle both set it) it can leak into the accessibility tree. One attribute.
> 
>   4. Spacing change to carry over
> 
>   [Automattic/vip-dashboard#5441 ](https://github.com/Automattic/vip-dashboard/issues/5441)changes the card's mb: 5 → mb: 6 (24px → 32px), from a design QA pass on the APM page. It'll need folding into this port,
>   otherwise it reverts as soon as the page switches to the design system component.
> 
>   ---
>   Items 1–3 are independent, so they can land separately if any turn out to be contentious. Item 4 is the only one with a merge-order dependency.

PR #711, opened 2026-09-14
Make the DropdownItem reset actually apply

donaghokeeffe:
## Description

`DropdownItem`'s style object opens with `unset: 'all'`. The property is `all` and the value is `unset` — as written this is not a CSS property, so the browser discards it. **The reset has never applied, anywhere, since it was added.**

The visible consequence is that the browser's default focus ring survives on menu items. Radix moves DOM focus onto whichever item is highlighted, including on pointer move, so the ring follows the mouse around the menu. Reported as a defect during QA more than once, and worked around locally by consumers.

Three commits, because correcting the typo on its own would have made things worse:

Dropdown-focus-border-reset

### 1. Make the reset apply

One character. `all: 'unset'`.

### 2. Put the keyboard focus ring back

`all: unset` clears the focus ring for keyboard navigation as well as the mouse — it cannot tell them apart. On its own that trades a bug annoying mouse users for one that materially hurts keyboard users: the only remaining indication of the highlighted item would be a 2% background tint, **1.04:1** against the menu, well under the **3:1** a focus indicator needs.

So the ring is restored deliberately, for keyboard only:

```js
'&:focus-visible:not(:hover)': { outline: 'revert' },
```

The `:not(:hover)` guard covers Radix focusing an item the mouse merely sits over, which is what made the ring track the pointer in the first place.

This is the pattern `EnvironmentSelectorV2Item`, `AppSwitcherItem` and an Overview menu in vip-dashboard already run in production — so those workarounds become genuinely redundant rather than merely duplicated.

### 3. Align the hover background with the environment switcher

Items and switcher rows sit in the same menus and tinted differently. Now both use `input.radio-box.background.hover`.

Deliberately **not** a focus indicator — at 1.11:1 it could not be one. The keyboard ring does that job; on hover this only reinforces the accent colour the label already takes, alongside the pointer cursor.

Worth flagging that the token is a radio-box one, which is the wrong domain for a menu item. It is what the switcher reaches for because there is no generic hover surface token. Raising separately rather than inventing one here.

## What still needs eyes

The risk in this change is entirely visual, and the test suite cannot see it — green CI here is necessary but nowhere near sufficient. The Storybook preview is the instrument.

**Four item types share the reset**, so all four change together:

| Component | Class |
|---|---|
| `DropdownItem` | `.vip-dropdown-menu-item` |
| `DropdownCheckboxItem` | `.vip-dropdown-checkbox-item` |
| `DropdownRadioItem` | `.vip-dropdown-radio-item` |
| `DropdownSubTrigger` | `.vip-dropdown-sub-trigger` |

`FilterDropdown` builds on `Dropdown.CheckboxItem`, so it is affected too.

Look for:

- **Typography** — items inherit font family, size and weight from their container. `unset` resolves inherited properties to `inherit`, so these should survive. This is the assumption most likely to be wrong and the likeliest place for a regression to hide.
- Backgrounds and borders across hover, highlight and disabled states.
- Anything the theme's `styles.root` or global CSS was contributing to an item.
- Sub-menu triggers, which merge the reset with their own open-state styles.

vip-dashboard renders dropdown or filter items in **28 files**, not covered by this preview. The full list is in the review thread and wants a pass in the consuming app before release.

## Checklist

- [x] This PR has good automated test coverage — existing suites pass; see the caveat above about what they can prove
- [ ] The storybook for the component has been updated — N/A, no API change

## Steps to Test

1. Open the Storybook preview on this PR.
1. Visit the **Dropdown** and **FilterDropdown** stories.
1. **Mouse:** hover down a menu. The focus ring should no longer track the pointer. Items should tint and the label take the accent colour.
1. **Keyboard:** Tab in and arrow through. Each highlighted item **should** carry a visible focus ring.
1. Compare all four item types against the same stories on trunk, looking for anything that shifted rather than only at the focus ring.


donaghokeeffe:
## vip-dashboard surfaces to check before release

Every file below renders a dropdown, checkbox, radio or sub-trigger item, so all are affected by this change. Grouped by where they appear in the product rather than by path, so they can be walked through in the browser.

### Global chrome — visible on nearly every page, so check these first

- Environment switcher in the sidebar — `EnvironmentSelector.tsx`, `EnvironmentSelectorV2.tsx`, `EnvironmentSelectorV2Item.tsx`
- Application switcher in the nav — `Navigation/AppSwitcher/AppSwitcher.tsx`, `AppSwitcherItem.tsx`
- User menu, top right — `PageNav/UserContent.tsx`
- Navigation item toggle — `NavigationItemToggle/NavigationItemToggle.tsx`

### Application Overview

- "More options" overflow menu — `Apps/Overview/AppHeader.tsx`
- Stat card customise menu (checkbox items, currently behind a VIP flag) — `Apps/Overview/CustomiseStatCardsMenu.tsx`
- `Apps/Overview/IntegrationsCard.tsx`
- `Apps/Overview/MetricsCardChart.tsx`
- `Apps/Overview/PremiumServicePromoCard.tsx`

### Integrations

- `Integrations/NetworkSiteTable.tsx`
- `Integrations/OrgApplicationRow.tsx`
- `Integrations/RdbSource/RdbSourceDataSourcesTable.tsx`
- `Integrations/Tollbit/TollbitFormRow.tsx`
- `views/Integrations/IntegrationRow.tsx`
- `views/IntegrationsCenter/Form/AddIntegrationDropdown.tsx`
- `views/IntegrationsCenter/IntegrationsList.tsx`

### Application views

- `Apps/MultiDirectionalSync/EnvironmentDropdown.tsx`
- `Apps/NetworkSites/NetworkSites.tsx`
- `Apps/Performance/VipApm/VipApm.tsx`
- `Apps/Plugins.tsx`

### User and organisation settings

- Personal access tokens, expiry dropdown — `User/Settings/PersonalAccessToken.tsx`
- Machine user access tokens — `User/Settings/MachineUserAccessToken.tsx`
- Organisation users list filter — `org/Users/UsersList/Filters.tsx` (this one is a `FilterDropdown`)

### Metrics

- `Health/Metrics/Overview/Issues.tsx`
- `Usage/Requests.tsx`

### Worth prioritising

Three of these already carry local workarounds for the focus ring this change fixes, so they are the most likely to show a double-up or a behaviour change: the environment switcher, the application switcher, and the Overview stat card customise menu.

The organisation users filter is the only `FilterDropdown` in the list, so it is the one place to confirm that path specifically.

luiztiago (approved):

